PR #21, opened 2019-06-04
NN-1953 add default values for new database columns

JamieDraperUK:


src/main/resources/db/migration/V4__add_iepwarn_case_note_id.sql
```diff
@@ -1,11 +1,11 @@
 ALTER TABLE  OFFENDER_ATTENDANCE
-ADD COLUMN CREATE_DATETIME  TIMESTAMP NOT NULL;
+ADD COLUMN CREATE_DATETIME  TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP;
 
 ALTER TABLE  OFFENDER_ATTENDANCE
-ADD COLUMN CREATE_USER_ID  VARCHAR( 32)    NOT NULL;
+ADD COLUMN CREATE_USER_ID  VARCHAR( 32) NOT NULL "SYSTEM";
```
petergphillips: missing `DEFAULT`?  Not sure it matters...

PR #26, opened 2019-06-07
Nn 1953 attendance put part 2

JamieDraperUK:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
```diff
@@ -77,35 +56,64 @@ public void updateAttendance(long id, UpdateAttendanceDto attendanceDto) throws
...
+    @Transactional
+    public void createAttendance(final CreateAttendanceDto attendanceDto) {
 
-    private void applyAttendanceWorkflow(String offenderNo, Attendance attendance) {
-         final var eventOutcome = nomisEventOutcomeMapper.getEventOutcome(
-                 attendance.getAbsentReason(),
-                 attendance.isAttended(),
-                 attendance.isPaid());
+        var attendance = attendanceRepository.save(toAttendance(attendanceDto));
+        entityManager.flush();
```
petergphillips: what's going on here then?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
```diff
@@ -77,35 +56,64 @@ public void updateAttendance(long id, UpdateAttendanceDto attendanceDto) throws
...
-         }
-     }
+        attendance.setAttended(attendanceDto.getAttended());
+        attendance.setPaid(attendanceDto.getPaid());
+        attendance.setAbsentReason(attendanceDto.getAbsentReason());
+        attendance.setComments(attendanceDto.getComments());
+
+        applyAttendanceWorkflow(attendance);
+
+        attendanceRepository.save(attendance);
```
andrewrlee: is there dirty checking/transactional advice which will mean that changes get persisted on commit anyway?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
```diff
@@ -77,35 +56,64 @@ public void updateAttendance(long id, UpdateAttendanceDto attendanceDto) throws
...
+                attendance.getAbsentReason(),
+                attendance.getAttended(),
+                attendance.getPaid());
+
+        nomisService.putAttendance(attendance.getOffenderBookingId(), attendance.getEventId(), eventOutcome);
+
+        addCaseNoteIfRequired(attendance).ifPresent(attendance::setCaseNoteId);
+    }
+
+     private Optional<Long> addCaseNoteIfRequired(Attendance attendance) {
```
andrewrlee: Doesn't seem like you need to return anything here? the return value is not being used?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
```diff
@@ -77,35 +56,64 @@ public void updateAttendance(long id, UpdateAttendanceDto attendanceDto) throws
...
+        attendance.setAbsentReason(attendanceDto.getAbsentReason());
+        attendance.setComments(attendanceDto.getComments());
+
+        applyAttendanceWorkflow(attendance);
+
+        attendanceRepository.save(attendance);
+    }
+
+
+    private void applyAttendanceWorkflow(Attendance attendance) {
```
andrewrlee: missing final... :-D 
maybe something like `publishXXX`/`publishToNomis` might be a bit more descriptive? 
(If that's what this is doing)

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
JamieDraperUK: The case note id is used on line 99.  I suspect I've misunderstood you though.

andrewrlee: oh no - I must have just misread..

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
JamieDraperUK: Yeah, there is, but wasn't sure how to trigger it in my test so called it explicitly. What's the best practice on this?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
JamieDraperUK: I like it!

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
JamieDraperUK: The idea is that we do the main save, hopefully failing fast so we don't proceed with the API calls. If everything is ok, crack on with the API call's then do a final update to add the case note id. 

I'm not really convinced it's giving us much though because if it was going to fail it would for both updates unless the nature of errors is validation which should be resolved before getting to the service.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
```diff
@@ -77,35 +56,64 @@ public void updateAttendance(long id, UpdateAttendanceDto attendanceDto) throws
...
+        attendance.setAbsentReason(attendanceDto.getAbsentReason());
+        attendance.setComments(attendanceDto.getComments());
+
+        publishAttendanceAndIEPWarning(attendance);
+
+        attendanceRepository.save(attendance);
+    }
+
+
+    private void publishAttendanceAndIEPWarning(final Attendance attendance) {
```
JamieDraperUK: Not fully convinced on the name as the IEP warning is optional.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
andrewrlee: Not too sure - I think it's [pretty difficult](https://blog.apnic.net/2018/02/08/testing-springs-transactional/) to unit test without setting up mock transaction managers and associated cruft.

Could argue the transactionality is a cross cutting concern that doesn't need explicitly testing in a unit testing and is more the remit for integration tests.
Are you currently verifying the save?

JamieDraperUK: Yeah,  save is verified in the AttendanceServiceTes.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
petergphillips: this isn't going to do anything though.  You are inside a transaction.  The only way to commit at this point is to end the transaction and start a new one.

PR #31, opened 2019-06-12
DT-50 Reduce to trace the request logging, reorder Filters, add W3C s…

Mjwillis:
…upport for distributed tracing. Add correlationId filter

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/utils/CorrelationMdcFilter.java
+++ b/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/utils/CorrelationMdcFilter.java
@@ -1,0 +1,53 @@
+package uk.gov.justice.digital.hmpps.whereabouts.utils;
+
+import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/utils/CorrelationMdcFilter.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public void doFilter(final ServletRequest request, final ServletResponse response, final FilterChain chain)
+            throws IOException, ServletException {
+
+        final var correlationIdOptional = Optional.ofNullable(((HttpServletRequest) request).getHeader(CORRELATION_ID_HEADER));
+
+        try {
+            MDC.put(CORRELATION_ID_HEADER, correlationIdOptional.orElseGet(mdcUtility::generateUUID));
```
andrewrlee: Do we have a way of then propagating this down to the elite 2 api calls?

PR #32, opened 2019-06-13
NN-1981 case note amendments

JamieDraperUK:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
```diff
@@ -130,4 +154,11 @@ private AttendanceDto toAttendanceDto(Attendance attendanceData) {
                 .caseNoteId(attendanceData.getCaseNoteId())
                 .build();
      }
+
+    private String getIepRescindedReason(final UpdateAttendanceDto newAttendance) {
+        if (newAttendance.getAttended())
+            return "IEP rescinded: attended";
+
+        return "IEP rescinded: " + newAttendance.getAbsentReason().toString();
+    }
```
Nimphal: Could this be something like:
```
return "IEP rescinded: " + (newAttendance.getAttended() ? "attended" : newAttendance.getAbsentReason().toString())
``` not sure if it's sufficiently readable, but reduces the string duplication of "IEP rescinded: "

Comment (JamieDraperUK) [reply src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java]: I like it!

File: src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java [hunk 2]
```diff
@@ -49,46 +48,71 @@ public AbsentReasonsDto getAbsenceReasons() {
...
         var attendance = attendanceRepository.save(toAttendance(attendanceDto));
-        publishAttendanceAndIEPWarning(attendance);
+
+        postNomisAttendance(attendance);
+        postIEPWarningIfRequired(
+                attendance.getOffenderBookingId(),
+                attendance.getCaseNoteId(),
+                attendance.getAbsentReason(),
+                attendance.getComments()
+        ).ifPresent(attendance::setCaseNoteId);
```
Nimphal: Just to check I've understood this correctly - postIEPWarningIfRequired could return a CaseNote Id or it could return empty. If it does return a CaseNote Id, we set that on the attendance. 

Would id matter if we tried to set an empty case note id on the attendance?

JamieDraperUK: That's correct, however, if the case note id is empty nothing will happen. The ifPresent only executes if there is a value.

Nimphal: Yeap, gotcha. Updating is quite an involved method now (and growing with my work), but I dont't have better ideas right now :D

PR #34, opened 2019-06-14
NN-1982 return modified by information

JamieDraperUK:


src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/AttendanceIntegrationTest.kt
```diff
@@ -554,4 +555,53 @@ class AttendanceIntegrationTest : IntegrationTest () {
 
         )
     }
+
+    @Test
+    fun `should return modified by and modified on`() {
+
```
Nimphal: Seeing as this is an integration test and that we have other tests in that file that also test the AttendanceDto, is it worth combining them all in one `should return expected serialised object` test that checks for everything we'd expect, especially if it's non-conditional? I can see it making sense to have multiple tests for locked as that has logic behind it, but for the others, we're just checking that the data is in the object.

JamieDraperUK: It's not just testing that the dto is correct, it also tests the following:-
  *@EntityListeners(AuditingEntityListener.class) is setup correctly
  * Modified only gets updated on updates
  * The correct modified time and modified gets returned correctly through the layers.

Nimphal: Ah, I see, did not notice that from the PR diff!

PR #35, opened 2019-06-18
NN-2009 add logging, tidy up iep logic and sentence case absent reasons

JamieDraperUK:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
```diff
@@ -68,40 +72,51 @@ public void updateAttendance(final long id, final UpdateAttendanceDto newAttenda
...
                         AbsentReason.getIepTriggers().contains(newAttendanceDetails.getAbsentReason());
 
         final var shouldRevokePreviousIEPWarning = attendance.getCaseNoteId() != null && !shouldTriggerIEPWarning;
 
         if (shouldRevokePreviousIEPWarning) {
-            final var rescindedReason = "IEP rescinded: " + (newAttendanceDetails.getAttended() ?
-                    "attended" : newAttendanceDetails.getAbsentReason().toString());
+            final var rescindedReason = "IEP rescinded: " + (
+                    newAttendanceDetails.getAttended() ? "attended" :
+                            capitalize(lowerCase(join(splitByCharacterTypeCamelCase(newAttendanceDetails.getAbsentReason().toString()), ' ')))
```
JamieDraperUK: Feels a bit long-winded, is there a better way?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
```diff
@@ -68,40 +72,51 @@ public void updateAttendance(final long id, final UpdateAttendanceDto newAttenda
...
+            );
```
Nimphal: Could this be assigned to a variable like `formattedAbsentReason` or something like that so the logic isn't part of this if statement?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
Nimphal: There doesn't seem to be an easier way to convert a string to title case :/

PR #36, opened 2019-06-19
Nn 1976 push att comment and update case note text

Nimphal:
This does two things:
1. Pushes the attendance comment to NOMIS (current only stored in this API's database)
2. Prepends a case note with the absence reason to provide more context for the IEP warning.


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/EventOutcome.java
```diff
@@ -4,10 +4,11 @@
 import lombok.AllArgsConstructor;
 import lombok.Data;
 
-@Data
+@Data()
```
JamieDraperUK: Are the parentheses required?

Nimphal: No, part of me debugging. The Attendance model has them so I put them in just in case.

PR #37, opened 2019-06-19
Nn 1976 format absent reason in case note

Nimphal:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/utils/AbsentReasonFormatter.java
```diff
@@ -0,0 +1,11 @@
+package uk.gov.justice.digital.hmpps.whereabouts.utils;
+
+import static org.apache.commons.lang3.StringUtils.*;
+
+public class AbsentReasonFormatter {
+
+    public static String titlecase( final String reason ) {
```
petergphillips: Since this works only on absent reasons then is should pass in the enum `AbsentReason` instead

petergphillips: but all the absent reasons are known in advance, so why not add them to `AbsentReason` instead and you won't need to dynamically work them out?

PR #39, opened 2019-06-21
NN-2009: Remove devtools as apparently cause of problem

petergphillips:


build.gradle
```diff
@@ -78,13 +78,12 @@ dependencies {
...
     compileOnly 'org.projectlombok:lombok:1.18.8'
 
-    runtime 'org.hsqldb:hsqldb:2.4.1'
+    runtime 'com.h2database:h2:1.4.199'
     runtime 'org.flywaydb:flyway-core:5.2.4'
     runtime 'org.postgresql:postgresql:42.2.6'
 
     implementation 'org.hibernate:hibernate-java8'
 
-    implementation 'org.springframework.boot:spring-boot-devtools'
```
petergphillips: See https://github.com/spring-projects/spring-boot/issues/16512

PR #41, opened 2019-06-25
NN-2017 IEP reinstated case amendment

Robc24:
Added ability to create case note amendment to show IEP being reinstated after it has previously been rescinded.


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
```diff
@@ -110,14 +112,30 @@ public void updateAttendance(final long id, final UpdateAttendanceDto newAttenda
                     rescindedReason);
 
             return Optional.empty();
-        } else {
+        } else if (shouldReinstatePreviousIEPWarning) {
+            final var formattedAbsentReason = newAttendanceDetails.getAbsentReason() != null ?
```
JamieDraperUK: Could move formattedAbsentReason above the if block to remove the duplication.

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceServiceTest.kt
```diff
@@ -769,6 +769,34 @@ class AttendanceServiceTest {
                 .putCaseNoteAmendment(1, 1, "IEP rescinded: attended")
     }
 
+    @Test
+    fun `should post case note amendment going from paid attendance to unpaid absent refused`() {
```
JamieDraperUK: I think this tests the scenario:
  -> unpaid (IEP Warning) -> paid attendance (IEP rescinded) -> unpiad (IEP warning reinstated)

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceServiceTest.kt
```diff
@@ -769,6 +769,34 @@ class AttendanceServiceTest {
...
+                .absentReason(AbsentReason.UnacceptableAbsence)
+                .comments("Unacceptable absence - No show.")
+                .build())
+
+        verify(nomisService)
+                .putCaseNoteAmendment(1, 1, "IEP reinstated: Unacceptable absence")
+    }
+
     @Test
     fun `should not post a case note amendment going from paid attendance to unpaid absent refused`() {
```
JamieDraperUK: This description is saying the case note amendment should 'not' be created, however, the never() has been removed making it inconsistent.

JamieDraperUK: The previous test case is still valid, might be worth reverting this change.

JamieDraperUK: Ignore the above comments, thought the code below was related to 'should not post a case note amendment going from paid attendance to unpaid absent refused'

PR #47, opened 2019-06-27
NN-2011 include oauth url

JamieDraperUK:


helm_deploy/values-dev.yaml
```diff
@@ -22,5 +22,6 @@ ingress:
 env:
   JWT_PUBLIC_KEY: "LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0NCk1JSUJJakFOQmdrcWhraUc5dzBCQVFFRkFBT0NBUThBTUlJQkNnS0NBUUVBemN4N1lidzJCQVd2N1NzRUlVR1ANCml5SlNJRGd0cUJ4MTlUd203dUkzVFg1elJjclpsVXhjRW1QSkdVZ3krRDJKSWhWbHFtZXF3R1YyQ05PcVpCZ0cNCmo4ZUpHQTU5aUlUemU4ZG1SSk5JYzdsNmxESmc5RE5KVk9pTHFVbFpGRENJcXplSTYzb3E2dWhjY2c1RFBpVE4NCnFPR1pjOHVwTit3NWRack52KzJHTGd4SzJwZXRNVS9CaFllV2YzS3ZZSU8xNnYxdXZuRk9HTzEzSG9XdTVCdGQNClN0L1RnY2xGaFZMRWRHdzdYYmlZSG5OWklkaHdhTlFpWWdtWG1qWlZkTXlDUERNbzEwTGtWMXAzVXkxNXBNTXgNClVwc2xKYU8wNlZIYXJtY3ZWYzNleHg5NlpHTjE2T2U4ZWZoeG5Rdmhzd0ZrbXlYT25sSForNDI1MnJHcHlKTG8NCmx3SURBUUFCDQotLS0tLUVORCBQVUJMSUMgS0VZLS0tLS0="
   ELITE2API_ENDPOINT_URL: "https://gateway.t3.nomis-api.hmpps.dsd.io/elite2api/"
+  OAUTH_ENDPOINT_URL: "https://gateway.t3.nomis-api.hmpps.dsd.io/auth"
```
petergphillips: this is going to get confusing - does the URL end in a `/` or doesn't it?  Worth making consistent with elite2 url above - other apps all end in a slash (although I would prefer it if none of them did - I think I've had that conversation with @andrewrlee before)

src/main/resources/application.properties
```diff
@@ -14,6 +15,8 @@ logging.uris.exclude.regex=/health|/info|/ping|/swagger.*|/webjars/.*|/v2/api-do
 azure.application-insights.web.enable-W3C=true
 azure.application-insights.web.enable-W3C-backcompat-mode=true
 
+
+oauth.root.url = http://localhost:9090/auth
 whereabouts.api.client.id=whereabouts-api-client
 whereabouts.api.client.secret=clientsecret
-nomis.auth.url=http://localhost:9090/auth/oauth/token
+ouath.token.url=${oauth.root.url}/oauth/token
```
petergphillips: these shouldn't be in both properties and yaml files, shouldn't need properties file at all...

helm_deploy/values-dev.yaml
JamieDraperUK: Looking at the elite url code it adds a double forward slash at the moment :( I should just fix it.

src/main/resources/application.properties
```diff
@@ -14,6 +15,8 @@ logging.uris.exclude.regex=/health|/info|/ping|/swagger.*|/webjars/.*|/v2/api-do
 azure.application-insights.web.enable-W3C=true
 azure.application-insights.web.enable-W3C-backcompat-mode=true
 
+
+oauth.root.url = http://localhost:9090/auth
```
petergphillips: in case I forget to mention it later, whereabouts should have a health indicator for auth too (that calls ping)

PR #52, opened 2019-06-28
NN-2011 add missing values to postgres profile

JamieDraperUK:


src/main/resources/application-postgres.properties
```diff
@@ -5,4 +5,8 @@ spring.jpa.properties.hibernate.jdbc.lob.non_contextual_creation=true
 
 elite2.uri.root=${ELITE2API_ENDPOINT_URL}
 oauth.root.uri=${OAUTH_ENDPOINT_URL}
+
+whereabouts.api.client.id=${OAUTH_CLIENT_ID}
```
petergphillips: @JamieDraperUK need to discuss this property lark on Tuesday.  The idea of the environment variables is that they are automatically converted into spring properties so you shouldn't need any of this mapping at all - i.e. defining an environment variable of `OAUTH_CLIENT_ID` will automatically create a spring property `oauth.client.id` at runtime.  This can then be used directly as is.

PR #54, opened 2019-07-02
NN-2011 add test for oauth, fix env variables

JamieDraperUK:
- Tidy up URLs
- Use spring environment variables directly
- Add test to ensure oauth gets called per request
- Add to the ReadMe
- Update docker-compose file


src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/AttendanceIntegrationTest.kt
```diff
@@ -617,4 +614,49 @@ class AttendanceIntegrationTest : IntegrationTest () {
...
 
     }
+
+    @Test
+    fun `should request a new auth token for each new incoming request`() {
+        val bookingId = 1
+        val activityId = 1L
+
+        elite2MockServer.stubFor(
+                WireMock.put(urlPathEqualTo("/api/bookings/$bookingId/activities/$activityId/attendance"))
```
JamieDraperUK: Going to move WireMock stubs to another file in later commits.

PR #56, opened 2019-07-08
NN-2041 add new endpoint to return attendance info for booking ids

JamieDraperUK:
booking ids


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/OffenderEventController.java
```diff
@@ -16,10 +18,7 @@
 
 @Api(tags = {"whereabouts"})
 
-@RestController
-@RequestMapping(
-        value="whereabouts",
-        produces = MediaType.APPLICATION_JSON_VALUE)
+@Deprecated
```
petergphillips: should have comment saying why deprecated and what replaced with etc.

PR #64, opened 2019-08-02
NN-2100 use date of last attendance for locking

JamieDraperUK:


src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/AttendanceIntegrationTest.kt
```diff
@@ -388,76 +388,10 @@ class AttendanceIntegrationTest : IntegrationTest () {
 
         assertThat(savedAttendance.id).isGreaterThan(0)
         assertThat(savedAttendance.createUserId).isEqualTo("ITAG_USER")
-        assertThat(savedAttendance.locked).isEqualTo(true)
+        assertThat(savedAttendance.locked).isEqualTo(false)
         assertThat(response.statusCodeValue).isEqualTo(201)
     }
 
-    @Test
```
JamieDraperUK: Didn't really want to delete these integration tests. The problem I've got is the logic is based on created and modified date and can't find a way to mock them. The saving grace, however, is that the locking logic has good unit test coverage.

PR #65, opened 2019-08-06
NN-2110 get attendance for scheduled activities

JamieDraperUK:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/AttendanceController.java
```diff
@@ -112,6 +112,17 @@ public AttendanceDto postAttendance(
 
     }
 
+    @GetMapping("/{prison}/attendance-for-scheduled-activities")
+    @ApiOperation(value = "Return a set of attendance details for all offenders that have scheduled activity",
+        response = AttendanceDto.class,
```
andrewrlee: need to put the [response container](http://docs.swagger.io/swagger-core/v1.5.0/apidocs/io/swagger/annotations/ApiOperation.html#responseContainer()) too?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/NomisService.java
```diff
@@ -58,5 +65,23 @@ public void putCaseNoteAmendment(final long bookingId, final long caseNoteId, fi
...
+        final var response = restTemplate.exchange(url, HttpMethod.GET, null, responseType, prisonId, date, period);
+        final var body = response.getBody();
+
+        if (body == null)
+            return Collections.emptySet();
+
+        return body
+                .stream()
+                .filter(entry -> entry.containsKey("bookingId"))
+                .map(entry -> Long.valueOf(entry.get("bookingId").toString()))
```
andrewrlee: should this be returning the booking ids? - not scheduled activity ids or something?  
(or is it a different type of booking? - like can you be booked on an activity?)

JamieDraperUK: Yeah, that's correct. Only the booking ids are required to pull the attendance information.

petergphillips: would be better to indicate that in the method name e.g. `getBookingIdsForScheduleActivities`

JamieDraperUK: Ah, crap yes!

PR #69, opened 2019-08-29
DT-180: Upgrade libs and improve health check

petergphillips:


helm_deploy/whereabouts-api/templates/deployment.yaml
```diff
@@ -34,7 +34,7 @@ spec:
               protocol: TCP
           livenessProbe:
             httpGet:
-              path: /health
+              path: /ping
```
petergphillips: once service is up stop other services failing taking down this service (cascading failures)

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/health/HealthCheck.java
```diff
@@ -1,39 +1,26 @@
...
-
     @Override
     public Health health() {
         try {
             final var responseEntity = this.restTemplate.getForEntity("/ping", String.class);
-            return health(responseEntity.getStatusCode());
+            return Health.up().withDetail("HttpStatus", responseEntity.getStatusCode()).build();
         } catch (final RestClientException e) {
-            return health(HttpStatus.SERVICE_UNAVAILABLE);
+            return Health.down(e).build();
```
petergphillips: service is `DOWN` if can't connect to dependencies

PR #70, opened 2019-08-29
DT-180: Add in dependency on case notes service

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/config/RestTemplateConfiguration.java
```diff
@@ -99,12 +104,12 @@ public RestOperations getRestTemplate() {
...
         private final AuthenticationFacade authenticationFacade;
 
-        TokenRequestEnhancer(AuthenticationFacade authenticationFacade) {
+        TokenRequestEnhancer(final AuthenticationFacade authenticationFacade) {
             this.authenticationFacade = authenticationFacade;
         }
 
         @Override
-        public void enhance(AccessTokenRequest request, OAuth2ProtectedResourceDetails resource, MultiValueMap<String, String> form, HttpHeaders headers) {
+        public void enhance(final AccessTokenRequest request, final OAuth2ProtectedResourceDetails resource, final MultiValueMap<String, String> form, final HttpHeaders headers) {
```
JamieDraperUK: Do you need to make sure this is all working in request scope?

PR #77, opened 2019-09-17
NN-2177 fix update attendance issue

JamieDraperUK:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/Elite2ApiService.java
```diff
@@ -28,15 +28,12 @@ public void putAttendance(final Long bookingId, final long activityId, final Eve
...
-        final var performance = eventOutcome.getPerformance();
-        final var comments = eventOutcome.getOutcomeComment();
-
-        final var body = Map.of(
-                "bookingActivities", bookingActivities,
-                "eventOutcome", eventOutcome.getEventOutcome(),
-                "performance", performance != null ? performance : "",
-                "outcomeComment", comments != null ? comments : ""
-        );
+        final var body = new HashMap<String, Object>();
```
JamieDraperUK: Couldn't use mapOf because it created an immutable map which doesn't allow nulls for values.

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/common/LocalDateDeserializer.kt
```diff
@@ -10,7 +10,9 @@ private class LocalDateDeserializer : JsonDeserializer<LocalDate> {
     }
 }
 
-
 fun getGson() : Gson {
-    return GsonBuilder().registerTypeAdapter(LocalDate::class.java, LocalDateDeserializer()).create()
+    return GsonBuilder()
+            .serializeNulls()
```
JamieDraperUK: Allow assertions on null values.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/Elite2ApiService.java
```diff
@@ -28,15 +28,12 @@ public void putAttendance(final Long bookingId, final long activityId, final Eve
 
     public void putAttendanceForMultipleBookings(final Set<BookingActivity> bookingActivities, final EventOutcome eventOutcome) {
         final var url = "/bookings/activities/attendance";
-        final var performance = eventOutcome.getPerformance();
-        final var comments = eventOutcome.getOutcomeComment();
-
-        final var body = Map.of(
-                "bookingActivities", bookingActivities,
-                "eventOutcome", eventOutcome.getEventOutcome(),
-                "performance", performance != null ? performance : "",
```
JamieDraperUK: Sending an empty string caused the elite to throw a bad request because it couldn't find a performance value of '""'

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/Elite2ApiService.java
petergphillips: um, can't you leave the fields out of the map that could be null.  See https://stackoverflow.com/questions/45210398/why-does-map-of-not-allow-null-keys-and-values - it is considered a design error that `HashMap` ever allowed nulls in the first place...

PR #83, opened 2019-10-08
overide global custom-http-errors setting

mattops:


helm_deploy/whereabouts-api/templates/ingress.yaml
```diff
@@ -13,6 +13,7 @@ metadata:
   annotations:
     kubernetes.io/ingress.class: "nginx"
     {{ if .Values.ingress.enable_whitelist }}nginx.ingress.kubernetes.io/whitelist-source-range: {{ include "app.joinListWithComma" .Values.whitelist | quote }}{{ end }}
+    nginx.ingress.kubernetes.io/custom-http-errors: "418"
```
petergphillips: Why just override this one?

petergphillips: Or is this the one you don't want to override?

mattops: 418 is the HTTP "I'm a teapot" error.  This is the only way to override the globally set values.  It's just a dummy value.  So assuming we never intend to serve up a genuine 418 HTTP error response from our apps- we should be good.

PR #86, opened 2019-10-22
DT-320: added telemetry module to log in app insights the client ID a…

SimonMitchellMOJ:
…nd username of each API request


src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/PingEndpointIntegrationTest.kt
```diff
@@ -7,10 +7,12 @@ import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.boot.test.context.SpringBootTest
 import org.springframework.boot.test.web.client.TestRestTemplate
 import org.springframework.http.HttpStatus
+import org.springframework.test.context.ActiveProfiles
 import org.springframework.test.context.junit4.SpringRunner
 import org.springframework.util.MimeType
 
 @RunWith(SpringRunner::class)
+@ActiveProfiles("test")
```
andymarke: Did this need to be test profile? I would have thought it didn't require any jwt stuff

SimonMitchellMOJ: Without it we get **Failed to load ApplicationContext** because the JwtAuthenticationHelper.class could not resolve placeholder, the values are only held in the application-test.properties

PR #90, opened 2019-10-28
DT-334: Bump replicas, set surge and anti affinity

petergphillips:


helm_deploy/whereabouts-api/templates/deployment.yaml
```diff
@@ -25,6 +26,15 @@ spec:
...
     spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
+              - key: app
+                operator: In
+                values:
+                - {{ template "app.name" . }}
```
petergphillips: Note that we don't need to specify the namespace since from https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#inter-pod-affinity-and-anti-affinity:
> If omitted or empty, it defaults to the namespace of the pod where the affinity/anti-affinity definition appears.
which means that it will only try to arrange pods in the same namespace on different nodes

helm_deploy/whereabouts-api/templates/deployment.yaml
```diff
@@ -25,6 +26,15 @@ spec:
         app: {{ template "app.name" . }}
         release: {{ .Release.Name }}
     spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
```
petergphillips: `preferred` since if we use `required` then it won't deploy the other pods if it can't find a node

helm_deploy/values-preprod.yaml
```diff
@@ -1,7 +1,7 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-replicaCount: 1
+replicaCount: 4
```
petergphillips: discussions with @mattops - preprod should have same settings as prod

PR #92, opened 2019-10-30
Nn 2232 whereabouts stats

JamieDraperUK:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/model/Attendance.java
```diff
@@ -74,4 +72,241 @@
     @LastModifiedBy
     @Column(name = "MODIFY_USER_ID")
     private String modifyUserId;
```
JamieDraperUK: Needed to delombok the class to get it to work with the attendance stats service.

Nimphal: I wonder if this is best approach considering wider code experience. Are we firmly committed to Kotlin and is it worth losing Lombok which is widely used for it?

JamieDraperUK: The plan is to slowly convert the whole application into Kotlin, however, I'm trying to do it iteratively.

PR #98, opened 2019-11-04
NN-2194 change name on stats

JamieDraperUK:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceStatistics.kt
```diff
@@ -29,7 +29,7 @@ open class AttendanceStatistics(private val attendanceRepository: AttendanceRepo
         notRecorded = offendersScheduledForActivity.count { !attendanceBookingIds.contains(it) },
         paidReasons = PaidReasons(
             attended = attendances.count { it.attended },
-            acceptableAbsences = attendances.count { it.absentReason == AbsentReason.AcceptableAbsence },
+            acceptableAbsence = attendances.count { it.absentReason == AbsentReason.AcceptableAbsence },
```
JamieDraperUK: Change the name to match the enum for easier referencing by the frontends.

PR #100, opened 2019-11-08
NN-2251 return new dto containing absent and offender details to drive stats screen

JamieDraperUK:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/Elite2ApiService.java
```diff
@@ -68,7 +68,8 @@ public String getOffenderNoFromBookingId(final Long bookingId) {
     public List<Long> getBookingIdsForScheduleActivitiesByDateRange(final String prisonId, final TimePeriod period, final LocalDate fromDate, final LocalDate toDate) {
         final var url = "/schedules/{prisonId}/activities-by-date-range?fromDate={fromDate}&toDate={toDate}&timeSlot={period}";
 
-        final var responseType = new ParameterizedTypeReference<List<Map>>() {};
+        final var responseType = new ParameterizedTypeReference<List<Map>>() {
+        };
```
Nimphal: Definitely not a blocker but looks like your IDE has been a bit overzealous again.

JamieDraperUK: I need to look into it but I suspect that's the standard Java way to format code with curly braces.

brightonsbox: Concise braces in Java are usually considered acceptable: https://google.github.io/styleguide/javaguide.html#s4.1.3-braces-empty-blocks

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/dto/AttendancesResponse.kt
```diff
@@ -13,6 +13,7 @@ import lombok.NoArgsConstructor
 @Builder(toBuilder = true)
 @ApiModel(description = "Attendances response")
 data class AttendancesResponse(@ApiModelProperty(required = true, value = "Set of attendances")
-                               var attendances: Set<AttendanceDto>? = null
+                               var attendances: Set<AttendanceDto>? = null,
+                               var absences: Set<AbsenceDto>? = null
```
brightonsbox: Add another @ApiModelProperty for the absences variable?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.java
```diff
@@ -276,32 +276,69 @@ private Boolean isAttendanceLocked(final Attendance attendance) {
...
+                details.getComment(),
+                attendance.getComments(),
+                details.getCellLocation(),
+                details.getFirstName(),
+                details.getLastName()
+        );
+    }
+
+    private Predicate<? super OffenderDetails> findAttendance(final Long bookingId, final Long eventId, final TimePeriod period) {
+        return offender -> offender.getBookingId().equals(bookingId) &&
```
brightonsbox: Probably minor, but it looks like OffenderDetails has a nullable bookingId? Could this raise a NullPointerException if it's not there? Same with the eventId below.

JamieDraperUK: Good spot, bookingId is fine, however eventId, at least in dev is sometimes null due to crappy data. I'll fix it anyways just to be safe.

PR #106, opened 2020-01-06
Nn 2342 throw error if attendance exists

Nimphal:
NN-2342: Throw error if user tries to create an attendance when one already exists


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/AttendanceController.kt
```diff
@@ -30,9 +31,14 @@ class AttendanceController(private val attendanceService: AttendanceService) {
       @ApiParam(value = "Attendance details", required = true)
       @RequestBody
       @Valid
-      attendance: CreateAttendanceDto): AttendanceDto {
+      attendance: CreateAttendanceDto): ResponseEntity<Any> {
+    val createdAttendance: AttendanceDto =  try {
+      attendanceService.createAttendance(attendance)
+    } catch (e: AttendanceExists) {
+      return ResponseEntity.badRequest().build()
```
JamieDraperUK: Should we return a http 409?

Nimphal: We definitely should, good catch!

JamieDraperUK: Might be worth returning an error message too.

Nimphal: 👍

PR #119, opened 2020-02-06
NN-2380 Integrate with the offender events queue and respond to delet…

JamieDraperUK:
…e events


docker-compose.yaml
```diff
@@ -37,3 +37,23 @@ services:
       - SERVER_PORT=8080
       - SPRING_PROFILES_ACTIVE=dev
       - ELITE2_API_BASE_URL=http://elite2-api:8080
+
+  localstack:
+    image: localstack/localstack
```
petergphillips: need to pin to 10.4 otherwise queue health with localstack doesn't work

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/config/JmsConfig.kt
```diff
@@ -0,0 +1,64 @@
...
+@Configuration
+@EnableJms
+@ConditionalOnProperty(name = ["sqs.provider"])
+open class JmsConfig {
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  @Bean
+  open fun jmsListenerContainerFactory(awsSqs: AmazonSQS?): DefaultJmsListenerContainerFactory {
```
petergphillips: `AmazonSQS` doesn't need to be nullable

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/config/JmsConfig.kt
```diff
@@ -0,0 +1,64 @@
...
+    val factory = DefaultJmsListenerContainerFactory()
+    factory.setConnectionFactory(SQSConnectionFactory(ProviderConfiguration(), awsSqs))
+    factory.setDestinationResolver(DynamicDestinationResolver())
+    factory.setConcurrency("3-10")
```
petergphillips: might be better to set concurrency to 1 - that will give 4 listeners in live anyway

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/EventListener.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.jms.annotation.JmsListener
+import org.springframework.stereotype.Service
+
+@Service
+open class EventListener(@Qualifier("attendanceServiceAppScope") private val attendanceService: AttendanceService) {
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+    val gson: Gson = GsonBuilder().create()
```
petergphillips: should be able to just inject gson instead

JamieDraperUK: Yep, no worries. It's copied from offender case notes.

src/main/resources/application-sqs.properties
```diff
@@ -0,0 +1,6 @@
+sqs.provider=aws
```
petergphillips: would be better to just put this in application.yml as the defaults then you won't need this profile

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/services/EventListenerTest.kt
```diff
@@ -0,0 +1,28 @@
...
+import com.nhaarman.mockito_kotlin.mock
+import com.nhaarman.mockito_kotlin.verify
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.junit.MockitoJUnitRunner
+import wiremock.org.apache.commons.io.IOUtils
+import java.io.IOException
+import java.nio.charset.StandardCharsets
+
+@RunWith(MockitoJUnitRunner::class)
```
petergphillips: don't need this annotation

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/EventListener.kt
JamieDraperUK: Although, not sure we need to inject it, thinking about it. Unless we have a specific reason we want the ability to swap out implementations?

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/services/EventListenerTest.kt
```diff
@@ -0,0 +1,28 @@
...
+class EventListenerTest {
+  private val attendanceService: AttendanceService = mock()
+
+  @Test
+  @Throws(IOException::class)
```
petergphillips: don't need any of these `throws`

src/main/resources/application-localstack.yml
```diff
@@ -0,0 +1,16 @@
+spring:
+  profiles:
+    include: dev
+
+sns:
```
petergphillips: why the `sns` configuration here?

src/main/resources/application-sqs.properties
```diff
@@ -0,0 +1,6 @@
...
+sqs.aws.access.key.id=
+sqs.aws.secret.access.key=
+sqs.queue.name=
+sqs.region=eu-west-2
+sqs.embedded=false
```
petergphillips: redundant property

src/main/resources/application-localstack.yml
JamieDraperUK: It shouldn't be. Left over from coping the pattern from offender-case-notes.

helm_deploy/whereabouts-api/templates/_envs.tpl
```diff
@@ -63,4 +63,41 @@ env:
       secretKeyRef:
         name: dps-rds-instance-output
         key: rds_instance_endpoint
+
+    - name: SQS_AWS_ACCESS_KEY_ID
+      valueFrom:
+        secretKeyRef:
+          name: ocn-events-sqs-instance-output
```
petergphillips: these all look wrong - you're grabbing the offender case note event output?  Should be grabbing the whereabouts output and configuring it in cloud platform environments

helm_deploy/whereabouts-api/templates/_envs.tpl
```diff
@@ -8,7 +8,7 @@ env:
     value: "{{ .Values.image.port }}"
 
   - name: SPRING_PROFILES_ACTIVE
-    value: "postgres,logstash"
+    value: "postgres,logstash,sqs"
```
petergphillips: prefer not to have the `sqs` profile

helm_deploy/whereabouts-api/templates/_envs.tpl
JamieDraperUK: This follows offender-case-notes and there doesn't seem to be any references to sqs modules in the cloud-platform-project.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/EventListener.kt
petergphillips: well it saves having to build another `Gson` - there is one available already so worth just using it - see https://github.com/ministryofjustice/keyworker-api/blob/master/src/main/java/uk/gov/justice/digital/hmpps/keyworker/events/EventListener.kt

helm_deploy/whereabouts-api/templates/_envs.tpl
petergphillips: https://github.com/ministryofjustice/cloud-platform-environments/tree/master/namespaces/live-1.cloud-platform.service.justice.gov.uk/offender-events-dev/resources

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.kt
```diff
@@ -215,6 +215,16 @@ open class AttendanceService(
         .collect(Collectors.toSet())
   }
 
+  @Transactional
+  open fun deleteAttendances(offenderNo: String) {
+    val bookingId = elite2ApiService.getOffenderBookingId(offenderNo)
+    val attendances = attendanceRepository.findByBookingId(bookingId)
+
+    log.info("Deleting the following attendance records ${attendances.map { it.id }.joinToString(",")}")
```
petergphillips: One for the next PR - I think we agreed that we should raise a telemetry event for the deletes for an offender too, but can't remember the exact format off the top of my head.  The idea is then that we can query app insights to see that the delete has filtered down to all systems

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/health/QueueHleath.kt
```diff
@@ -0,0 +1,86 @@
...
+
+enum class QueueAttributes(val awsName: String, val healthName: String) {
+  MESSAGES_ON_QUEUE(ApproximateNumberOfMessages.toString(), "MessagesOnQueue"),
+  MESSAGES_IN_FLIGHT(ApproximateNumberOfMessagesNotVisible.toString(), "MessagesInFlight"),
+  MESSAGES_ON_DLQ(ApproximateNumberOfMessages.toString(), "MessagesOnDLQ")
+}
+
+@Component
+@ConditionalOnProperty("sqs.provider")
+open class QueueHealth(@Autowired @Qualifier("awsSqsClient") private val awsSqsClient: AmazonSQS,
```
petergphillips: tpyo of kotlin filemane

PR #126, opened 2020-02-13
DT-526 Handle missing JSON fields for null values

mikehalmamoj:
Handle missing fields in JSON returned from elite2.


src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/wiremock/Elite2MockServer.kt
```diff
@@ -104,12 +104,29 @@ class Elite2MockServer : WireMockRule(8999) {
...
+    val locationMap = mutableMapOf(
+        "agencyId" to this.agencyId,
+        "currentOccupancy" to "${this.currentOccupancy}",
+        "description" to this.description,
+        "locationId" to "${this.locationId}",
+        "locationPrefix" to this.locationPrefix,
+        "locationType" to this.locationType,
+        "operationalCapacity" to "${this.operationalCapacity}",
+        "internalLocationCode" to this.internalLocationCode)
+    if (this.userDescription != null) locationMap["userDescription"] = this.userDescription as String
```
mikehalmamoj: The safe cast didn't work here but I don't know why

mikehalmamoj: Because it's calling the getter!

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/LocationIntegrationTest.kt
```diff
@@ -79,7 +79,7 @@ class LocationIntegrationTest: IntegrationTest() {
           operationalCapacity = 0,
           userDescription = "",
           internalLocationCode = "",
-          locationUsage = "",
+          locationUsage = null,
```
petergphillips: you're calling the constructor with named parameters so don't need to include nulls in the constructor

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/wiremock/Elite2MockServer.kt
mikehalmamoj: Because even though you can safely call its assuming it is not null, when passed to another function its type is still nullable - hence the cast is required

PR #132, opened 2020-02-24
DT-370 jwk server

mikehalmamoj:
Use the new JWK server from Nomis Oauth2 to retrieve the public key used to verify access tokens.


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/config/ClientTrackingTelemetryModule.java
```diff
@@ -63,6 +64,15 @@ public void onBeginRequest(final ServletRequest req, final ServletResponse res)
 
     private Claims getClaimsFromJWT(final String token) throws ExpiredJwtException, GeneralSecurityException {
 
+        if (jwtPublicKey == null) {
```
mikehalmamoj: Went for basic caching locally as this is what Spring Security does.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/config/JwkClient.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.List;
+
+@Component
+public class JwkClient {
+
+    private RestTemplate restTemplate;
+    private String keyId;
+
+    public JwkClient(@Qualifier("oauthApiHealthRestTemplate") final RestTemplate oauthApiHealthRestTemplate,
+                     @Value("${jwt.jwk.key.id}") final String keyId) {
```
mikehalmamoj: Spring security takes the key id from the JWT header and ideally we would do the same here.  Unfortunately Spring uses a package access class JwtHeaderConverter which is not available to us so we'd have to create our own version...

The benefit of taking the kid from the JWT header is that there is no change required in the client when we add a new JWK set hence we can rotate keys by simply adding them to the Oauth server and restarting.

src/test/resources/application-test.yml
```diff
@@ -19,7 +19,7 @@ management.endpoint:
...
+token: eyJ0eXAiOiJKV1QiLCJraWQiOiJkcHMtY2xpZW50LWtleSIsImFsZyI6IlJTMjU2In0.eyJqdGkiOiI2ZGY0YzFiOC0zMzhmLTRjNzItODMwNC01Mjc4NmJiZjIzNzgiLCJpbnRlcm5hbFVzZXIiOnRydWUsInVzZXJfbmFtZSI6IklUQUdfVVNFUiIsImNsaWVudF9pZCI6ImVsaXRlMmFwaWNsaWVudCIsImF1dGhvcml0aWVzIjpbIlJPTEVfTElDRU5DRV9DQSIsIlJPTEVfS1dfQURNSU4iXSwic2NvcGUiOlsicmVhZCIsIndyaXRlIl0sImV4cCI6MjIxMzAwNzA3Nn0.CS11l9ZmpfV7x_tIHPJnnKFHBQEuqYqtsjRfEzJ3wh_ByGeDVwVHnN5PkLVMlWUYkpHO1XQOlG_oK6-8bGV5avwwNtRFX526dI6t6Smrl0pupgj3zVRxVmKhuQ3XfOKHRizF_m8NH9_FyfCJICheBK7FpGZgJbZQf61eweYU3oeED-TXsrU3CBPTzglxvf5XzZRh1xMyWeq_q8EwAB6sFWQsqSBjWR1FHiQMrSo7NcCtr7Yqefjy_C7LurCPMldxOiUqLOXOw_GEAd2PAliTjTuo0mb9TUUZPSepo5YCLRAahEbunw4ufTY5iHwlaMNcrr5fiBrqDDiF5YVPBh1aiA
```
mikehalmamoj: The new token was required to add a `kid` to the header.

PR #145, opened 2020-03-31
NN-2529: Map scheduled activity correctly

Nimphal:


Nimphal:
The service itself is covered in a test, it didn't catch this because the API response is mocked.

PR #146, opened 2020-04-01
NN-2538: Refactor stats controller to make it possible to return number of suspended schedules per period

Nimphal:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/Elite2ApiService.java
```diff
@@ -72,24 +72,6 @@ public String getOffenderNoFromBookingId(final Long bookingId) {
         return (String) Objects.requireNonNull(entity.getBody()).get("offenderNo");
     }
 
-    public List<Long> getBookingIdsForScheduleActivitiesByDateRange(final String prisonId, final TimePeriod period, final LocalDate fromDate, final LocalDate toDate) {
```
Nimphal, line 75: This is no longer used as we now need two types of info from the schedules - booking id and suspended flag

PR #147, opened 2020-04-03
Nn 2512 spring 5 upgrade

JamieDraperUK:
Will be converting all the recently added Java to Kotlin in a future PR.

Fairly standard upgrade to spring security 5 apart from the requirement to extend the auth token request in order to pass on the username for system to system calls (client cred grant type) 

In order to alter the auth token request you need to hook into an extension point during the wiring up of OAuth2AuthorizedClientManager. This bean can't run in request scope.  From what I can tell this is because it uses the reactive webflux type magic and throws a wobbly when annotated with @RequestScope. With it being wired up outside of the request scope meant I didn't have access to the current authentication object pulled from  SecurityContextHolder.getContext().getAuthentication().  

What I ended up doing was extending UserContextFilter and UserContext to store the current Authentication object. By the time the filter is executed it already has a valid Authentication object all wired up previously by the resource server and config.


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/config/ClientTrackingTelemetryModule.java
```diff
@@ -4,41 +4,37 @@
...
-import java.security.spec.X509EncodedKeySpec;
+import java.security.PublicKey;
+import java.text.ParseException;
 import java.util.Optional;
 
 @Slf4j
 @Configuration
 public class ClientTrackingTelemetryModule implements WebTelemetryModule, TelemetryModule {
-    private final String jwtPublicKey;
+    private final PublicKeySupplier publicKeySupplier;
```
mikehalmamoj, line 27: @petergphillips discovered that the nimbus SignedJWT doesn't verify the token and hence doesn't need a public key.  He made the change in keyworker-api - might be one to consider for a future PR.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/config/AttendanceServiceForAppScope.kt
```diff
@@ -15,7 +15,7 @@ import uk.gov.justice.digital.hmpps.whereabouts.services.NomisEventOutcomeMapper
 class AttendanceServiceForAppScope {
   @Bean(name = ["attendanceServiceAppScope"])
   fun getAttendanceServiceForAppScope(
-      @Qualifier("elite2ApiRestTemplateAppScope") restTemplate: OAuth2RestTemplate,
+      @Qualifier("elite2WebClientAppScope") restTemplate: WebClient,
```
mikehalmamoj, line 18: Forgot to rename this variable...

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/config/WebClientConfiguration.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    private WebClient getOAuthWebClient(final OAuth2AuthorizedClientManager authorizedClientManager, final WebClient.Builder builder, final String rootUri) {
+        final var oauth2Client = new ServletOAuth2AuthorizedClientExchangeFilterFunction(authorizedClientManager);
+        oauth2Client.setDefaultClientRegistrationId("elite2-api");
+
+        final var exchangeStrategies =
+                ExchangeStrategies.builder()
+                        .codecs(configurer -> configurer.defaultCodecs().maxInMemorySize(-1))
+                        .build();
```
mikehalmamoj, line 125: It later transpired that this only affects one of the many decoders that are used, and we had to set a Spring property to set the max size (keyorker-api PR #240).

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.kt
```diff
@@ -27,7 +27,7 @@ open class AttendanceService(
     val log: Logger = LoggerFactory.getLogger(this::class.java)
   }
 
-  open fun getAttendanceForEventLocation(prisonId: String?, eventLocationId: Long?, date: LocalDate?, period: TimePeriod?): Set<AttendanceDto> {
+  fun getAttendanceForEventLocation(prisonId: String?, eventLocationId: Long?, date: LocalDate?, period: TimePeriod?): Set<AttendanceDto> {
```
mikehalmamoj, line 30: I know it's not this change, but I was just wondering if these parameters are really nullable or if it was just because of a Java conversion?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/IEPWarningService.kt
```diff
@@ -65,6 +64,8 @@ open class IEPWarningService(private val caseNotesService: CaseNotesService, pri
           "IEP_WARN",  //"IEP Warning",
           modifiedTextWithReason,
           eventDate.atStartOfDay())
+
+      log.info("IEP Warning created for bookingId {}", bookingId)
```
Nimphal, line 68: good spot on moving the logging after it's actually done :D

PR #149, opened 2020-04-06
NN-2512 spring 5 fix env vars

JamieDraperUK:


src/main/resources/application.properties
```diff
@@ -3,15 +3,11 @@ info.app.version=1.0
 logging.uris.exclude.regex=/health|/info|/ping|/swagger.*|/webjars/.*|/v2/api-docs
 azure.application-insights.web.enable-W3C=true
 courts=Kingston upon Thames,Westminster,Southwark,City of London,Wimbledon
-
-elite2.uri.root=http://localhost:8080
-elite2.api.uri.root=${elite2.uri.root}/api
-
-auth.endpoint.url=http://localhost:9090/auth
+elite2api.endpoint.url=http://localhost:8080
```
petergphillips, line 6: ideally we shouldn't be using properties files, everything should be in the yaml files.  Properties files were for before the yaml was invented.  Yaml allows better grouping and nesting of properties.  Obviously don't change, but worth bearing in mind

src/main/resources/application.properties
```diff
@@ -3,15 +3,11 @@ info.app.version=1.0
...
+elite2.api.uri.root=${elite2api.endpoint.url}/api
+oauth.endpoint.url=http://localhost:9090/auth
 casenotes.endpoint.url=http://localhost:8083
-
-api.health-timeout-ms: 1000
-
+api.health-timeout-ms:1000
```
petergphillips, line 10: `:` syntax is for yaml, not properties

PR #161, opened 2020-04-14
NN-2558 upgrade packages

JamieDraperUK:


build.gradle
```diff
@@ -114,12 +114,13 @@ dependencies {
...
 
     implementation 'org.springframework.boot:spring-boot-starter-security'
+
     implementation 'org.springframework.boot:spring-boot-starter-oauth2-resource-server'
     implementation 'org.springframework.boot:spring-boot-starter-oauth2-client'
     implementation 'org.springframework.security:spring-security-oauth2-jose'
+
     implementation 'org.springframework:spring-webflux'
     implementation 'org.springframework.boot:spring-boot-starter-reactor-netty'
-    implementation 'com.nimbusds:nimbus-jose-jwt:8.10' // pinned due to CVE-2019-17195 in 7.8.1 brought in by spring-boot-starter-oauth2-resource-server-2.2.5.RELEASE
```
brightonsbox, line 122: Does this not need to be pinned any more? I thought 2.2.6 still pulls in an affected version of this?

build.gradle
```diff
@@ -7,11 +7,11 @@ plugins {
     id 'java'
     id 'org.jetbrains.kotlin.jvm' version '1.3.61'
     id 'groovy'
-    id 'org.springframework.boot' version '2.2.5.RELEASE'
+    id 'org.springframework.boot' version '2.2.6.RELEASE'
     id 'io.spring.dependency-management' version '1.0.9.RELEASE'
-    id 'org.owasp.dependencycheck' version '5.3.0'
+    id 'org.owasp.dependencycheck' version '5.3.2'
```
brightonsbox, line 12: Meh... there is a 5.3.2.1 version which is the latest with some minor fixes

build.gradle
JamieDraperUK: Just double checked and 2.2.6 pulls in com.nimbusds:nimbus-jose-jwt:8.8 so think we're fine.

brightonsbox: Ah cool, ok! Perhaps I was thinking of something else.

PR #165, opened 2020-04-22
NN-2429: Update attendance deletion to take booking Ids from new DELETE event message

Nimphal:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.kt
```diff
@@ -218,15 +218,15 @@ class AttendanceService(
...
+      val attendances = attendanceRepository.findByBookingId(bookingId)
+      log.info("Deleting the following attendance records ${attendances.map { it.id }.joinToString(",")}")
 
-    log.info("Deleting the following attendance records ${attendances.map { it.id }.joinToString(",")}")
+      attendanceRepository.deleteAll(attendances)
 
-    attendanceRepository.deleteAll(attendances)
-
-    telemetryClient.trackEvent("OffenderDelete", mapOf("offenderNo" to offenderNo, "count" to  attendances.size.toString()), null)
+      telemetryClient.trackEvent("OffenderDelete", mapOf("bookingId" to bookingId.toString(), "count" to  attendances.size.toString()), null)
```
brightonsbox, line 228: Sorry Nev, would it be possible to pass the offenderNo through too and keep the telemetry event as it was?  It's just going to be easier to search for in Application Insights if all of the events have the same structure and we just need to search by offenderNo

Nimphal: why are you apologising :D on it

PR #166, opened 2020-04-24
NN-2569 record attendance changes

JamieDraperUK:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.kt
```diff
@@ -230,6 +244,36 @@ class AttendanceService(
...
+      AttendanceChangeDto(
+          id = it.id!!,
+          attendanceId = it.attendanceId,
+          bookingId = attendance.bookingId,
+          eventId = attendance.eventId,
+          eventLocationId = attendance.eventLocationId,
+          changedFrom = it.changedFrom,
+          changedTo = it.changedTo,
+          changedOn = it?.createDateTime,
+          changedBy = it?.createUserId
```
Nimphal, line 272: this right here is pretty genious

JamieDraperUK: lol Thanks. However, I might end of changing some of it as I really should let JPA deal with the change->attendance relationship.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/model/AttendanceChangeValues.kt
```diff
@@ -0,0 +1,15 @@
+package uk.gov.justice.digital.hmpps.whereabouts.model
+
+
+enum class AttendanceChangeValues {
```
JamieDraperUK, line 4: What I'd really like to do with the whole attendance model is remove the absent reason and attended boolean combination and have an overall attendance status much like this enum.

Nimphal: That would make it less brittle 👍 I also like the companion object thingie (I may have just had a course on them) for AbsentReason, much neater!

PR #168, opened 2020-04-28
NN-2569: Add endpoint which returns attendances for set of booking ids over date range

Nimphal:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/AttendancesController.kt
```diff
@@ -88,6 +88,19 @@ class AttendancesController(private val attendanceService: AttendanceService) {
...
   }
 
+  @PostMapping("/{prison}/attendance-over-date-range")
+  @ApiOperation(value = "Returns set of attendance details for set of booking ids", response = AttendancesResponse::class, notes = "Request attendance details")
+  fun getAttendanceForBookingsOverDateRangeByPost(@ApiParam(value = "Prison id (LEI)") @PathVariable(name = "prison") prisonId: String,
+                                                  @ApiParam(value = "Date of event in format YYYY-MM-DD", required = true) @RequestParam(name = "fromDate") @DateTimeFormat(iso = DATE) fromDate: LocalDate,
+                                                  @ApiParam(value = "Date of event in format YYYY-MM-DD defaults to fromDate") @RequestParam(name = "toDate") @DateTimeFormat(iso = DATE) toDate: LocalDate?,
+                                                  @ApiParam(value = "Time period. Leave blank for AM + PM") @RequestParam(name = "period") period: TimePeriod?,
+                                                  @ApiParam(value = "Booking ids (bookings=1&bookings=2)", required = true) @RequestBody bookings: Set<Long>): AttendancesResponse {
+
```
JamieDraperUK, line 98: Meh, example might be a bit misleading.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.kt
```diff
@@ -53,6 +53,18 @@ class AttendanceService(
     return attendance.map { toAttendanceDto(it) }.toSet()
   }
 
+  fun getAttendanceForBookingsOverDateRange(prisonId: String?,  bookings: Set<Long?>?, fromDate: LocalDate, toDate: LocalDate?,
```
JamieDraperUK, line 56: Does bookings here have to be nullable? It's required and set nullable at the controller level.  Same for prison.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.kt
```diff
@@ -53,6 +53,18 @@ class AttendanceService(
...
+                                        period: TimePeriod?): Set<AttendanceDto> {
+
+    val periods = if (period == null) setOf(TimePeriod.AM, TimePeriod.PM) else setOf(period)
+    val endDate = toDate ?: fromDate
```
JamieDraperUK, line 60: 😎

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.kt
Nimphal: good point, have tightened up

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/AttendancesController.kt
Nimphal: Fixed, and updated on the other post route to match.

PR #169, opened 2020-04-28
Nn 2569 update attendances integration tests

Nimphal:


src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/AttendancesIntegrationTest.kt
```diff
@@ -152,17 +151,17 @@ class AttendancesIntegrationTest : IntegrationTest() {
...
-            createHeaderEntity(""),
-            AttendancesResponse::class.java,
-            LocalDate.of(2019, 10, 10),
-            TimePeriod.PM)
-
-    assertThat(response.statusCodeValue).isEqualTo(200)
-    assertThat(response.body?.attendances).extracting("bookingId").contains(1L, 2L)
+    webTestClient
+      .get()
+      .uri("/attendances/LEI?date={0}&period={1}&bookings=${1}&bookings=${2}", LocalDate.of(2019, 10, 10), TimePeriod.PM)
```
JamieDraperUK, line 156: Is it worth using uri builder pattern with queryParam here?

Nimphal: When I was looking at the docs, they sort of implied that you can use the builder if you "need full programmatic control". That seemed a bit excessive, so I moved back to this. But happy to change.

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/AttendancesIntegrationTest.kt
```diff
@@ -45,35 +45,30 @@ class AttendancesIntegrationTest : IntegrationTest() {
...
-                .attended(false)
-                .paid(false)
-                .bookingId(1)
-                .locked(false)
-                .build())
+    webTestClient
+            .get()
+            .uri {
+              it.path("/attendances/LEI/2")
+                      .queryParam("date", LocalDate.of(2019, 10, 10))
```
JamieDraperUK, line 52: This is good! Should use this for the other calls below.

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/AttendancesIntegrationTest.kt
JamieDraperUK: Happy either way but let's pick one and make everything else consistent.  I think it would look cleaner to use the builder here because you could just use:
`queryParam("bookings', setof(1,2,3,4,))`

Nimphal: Updated to uri builder everywhere :)

PR #200, opened 2020-08-26
NN-2885: Remove restTemplate in favor of webTestClient. And some more renaming

Nimphal:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/config/WebClientConfiguration.java
```diff
@@ -25,26 +25,26 @@
...
+    private String prisonApiRootUri;
+    private String prisonApiHealthRootUri;
     private String caseNotesRootUri;
     private String oauthRootUri;
 
     public WebClientConfiguration(
-            @Value("${elite2.api.uri.root}") final String elite2ApiRootUri,
-            @Value("${elite2api.endpoint.url}") final String eliteHealthRootUri,
+            @Value("${elite2.api.uri.root}") final String prisonApiRootUri,
+            @Value("${elite2api.endpoint.url}") final String prisonApiHealthRootUri,
```
Nimphal, line 35: Deliberately did not change properties as part of this PR.

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/AgencyIntegrationTest.kt
```diff
@@ -1,54 +1,128 @@
...
 
   @Test
   fun `location groups - exist in properties - should retrieve groups from properties`() {
     val agencyId = "MDI"
 
-    val response: ResponseEntity<String> =
-        restTemplate.exchange("/agencies/$agencyId/locations/groups", HttpMethod.GET, createHeaderEntity(""))
-
-    assertThatJsonFileAndStatus(response, 200, "MDI_location_groups.json");
+    webTestClient.get()
```
JamieDraperUK, line 28: I think it looks much neater with those json files. What do you think? Might be worth tweaking assertThatJsonFileAndStatus later on at some point if you agree.

Nimphal: I generally agree, and I tried to get it budge, but didn't succeed. I think this particular test is an extreme example, but it is likely we might have other long ones. Haven't deleted the JSON files for that very reason.

JamieDraperUK (approved):
LGTM

PR #210, opened 2020-10-29
Nn 3014 cell move delegation endpoint

JamieDraperUK:
Initial PR to paving the way to storing cell move reason comments, and to raise case notes. 

The idea is that prisonstaffhub will make a cell move via whereabouts, in doing so it'll delegate calls to prison-api and store custom metadata, raise case notes and whatever else we'll need to satisfy upcoming requirements.


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/CellService.kt
```diff
@@ -0,0 +1,16 @@
+package uk.gov.justice.digital.hmpps.whereabouts.services
+
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.whereabouts.dto.CellMoveDetails
+import uk.gov.justice.digital.hmpps.whereabouts.dto.CellMoveResult
+
+@Service
+class CellService(val prisonApiService: PrisonApiService) {
+  fun makeCellMove(cellMoveDetails: CellMoveDetails): CellMoveResult {
+    return prisonApiService.putCellMove(
```
JamieDraperUK, line 10: Doesn't do much at the moment but it will orchestrate the creation of case notes and store comments.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/CellService.kt
```diff
@@ -0,0 +1,16 @@
+package uk.gov.justice.digital.hmpps.whereabouts.services
+
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.whereabouts.dto.CellMoveDetails
+import uk.gov.justice.digital.hmpps.whereabouts.dto.CellMoveResult
+
+@Service
+class CellService(val prisonApiService: PrisonApiService) {
```
Nimphal, line 8: Mostly because the first thing that came to mind was phone service - could this be named `CellMoveService` ?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/CellController.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.whereabouts.dto.CellMoveDetails
+import uk.gov.justice.digital.hmpps.whereabouts.dto.ErrorResponse
+import uk.gov.justice.digital.hmpps.whereabouts.services.CellService
+import javax.validation.Valid
+
+@Api(tags = ["cell"])
+@RestController
+@RequestMapping(value = ["cell"], produces = [MediaType.APPLICATION_JSON_VALUE])
+class CellController {
```
Nimphal, line 21: Similarly here re: naming - maybe CellMoveController is a bit more descriptive?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/CellService.kt
JamieDraperUK: Ah yes, it could actually.

Nimphal (approved):
LGTM!

PR #214, opened 2020-11-11
NN-3020 tidy up attendance controller and tests

JamieDraperUK:


src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/IntegrationTest.kt
```diff
@@ -45,6 +45,13 @@ abstract class IntegrationTest {
     @JvmField
     internal val caseNotesMockServer = CaseNotesMockServer()
 
+    private var lastBookingId = 0L
+
+    fun getNextBookingId(): Long {
```
JamieDraperUK, line 50: Not the best but works for now. I want to move all none end to end type tests into controller tests going forward in the hope to reduce complexity.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/ControllerAdvice.kt
```diff
@@ -0,0 +1,161 @@
...
+          .builder()
+          .status(HttpStatus.INTERNAL_SERVER_ERROR.value())
+          .developerMessage(e.message)
+          .build()
+      )
+  }
+
+  @ExceptionHandler(AccessDeniedException::class)
+  fun handleException(e: AccessDeniedException?): ResponseEntity<ErrorResponse> {
+    log.debug("Forbidden (403) returned", e)
```
petergphillips, line 60: shouldn't really log `e` here, maybe just the message, otherwise appears in app insights in the `exceptions` tab.  Same with the rest of file, only log `e` if it is a server error

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/ControllerAdvice.kt
```diff
@@ -0,0 +1,161 @@
...
+        ErrorResponse
+          .builder()
+          .status(HttpStatus.NOT_FOUND.value())
+          .developerMessage(e.message)
+          .build()
+      )
+  }
+
+  companion object {
+    val log = LoggerFactory.getLogger(this::class.java)
```
petergphillips, line 159: missing `private`

petergphillips (approved):
LGTM

PR #223, opened 2020-11-27
DCS-698 Link existing video link appointments into video link bookings

bell-pa:
Create a `VideoLinkBooking` for every main `VideoLinkAppointment`.  Add to these bookings the corresponding pre/post VideoLinkAppointments thereby linking related pre, main and post appointments into bookings.
In prod there are about 15,000 main appointments, 2000 pre appointments and 800 post appointments.

The linking process is started by a POST to `/api/courts/appointment-linker`. 
An alternative might be to give the linker function a separate entry point and run it as a Kubernetes job.
This will probably be run just once in production.


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/VideoLinkAppointmentLinker.kt
```diff
@@ -0,0 +1,116 @@
...
+        .filter { it.eventSubType == "VLB" }
+      prisonAppointments.addAll(chunk)
+      offset += chunk.size
+    } while (chunk.isNotEmpty())
+
+    log.info("bookingId $bookingId: Retrieved ${prisonAppointments.size} VLB appointments from prison-api")
+
+    return prisonAppointments
+  }
+
```
bell-pa, line 83: `preAppointmentFinder` and `postAppointmentFinder` return (a) PRE/POST appointment contiguous with the supplied appointment. For pre appointments this means `pre.endTime == main.startTime` and for post `post.startTime == main.endTime`.  This may be too strict. 
If so then these two functions are where a better appointment linking algorithm should be defined.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/VideoLinkAppointmentLinker.kt
```diff
@@ -0,0 +1,116 @@
...
+typealias AppointmentFinder = (PrisonAppointment?) -> VideoLinkAppointment?
+
+@Service
+class VideoLinkAppointmentLinker(
+  private val videoLinkAppointmentRepository: VideoLinkAppointmentRepository,
+  private val videoLinkBookingRepository: VideoLinkBookingRepository,
+  private val prisonApiService: PrisonApiService
+) {
+
+  fun linkAppointments() {
```
bell-pa, line 22: This rate at which this runs will be limited by the speed at which prison-api can return appointments per bookingId.    There are ~ 4300 bookingIds for which to fetch and filter all appointments since 2019-01-01.  There's no throttling here - perhaps there should be?

src/main/resources/db/migration/V14__video_link_booking_unique_constraints.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE video_link_booking ADD UNIQUE (pre_appointment);
+ALTER TABLE video_link_booking ADD UNIQUE (main_appointment);
+ALTER TABLE video_link_booking ADD UNIQUE (post_appointment);
+
+ALTER TABLE video_link_appointment ADD UNIQUE (appointment_id);
```
bell-pa, line 5: Add an index on video_link_appointment(booking_id) ? 18,000 rows? 4300 distinct booking_ids?

src/main/resources/db/migration/V14__video_link_booking_unique_constraints.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE video_link_booking ADD UNIQUE (pre_appointment);
```
bell-pa, line 1: These constraints express properties of the data and add indexes on pre,main,post appointment. 
Do they help? Are they superfluous?

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/repository/AppointmentLinkerSupportTest.kt
```diff
@@ -0,0 +1,223 @@
...
+   * bookingId 2 has one linked MAIN, PRE and POST appointment, two unlinked PRE and two unlinked POST appointment.
+   * bookingId 3 has two unlinked MAIN appointments. bookingId 3 should be in the result.
+   * bookingId 4 has one unlinked PRE appointment. bookingId 4 should not be in the result.
+   * bookingId 5 has one unlinked POST appointment. bookingId 4 should not be in the result.
+   */
+
+  private fun insertTestData() {
+    var appointmentIdCounter = 100L
+
+    videoLinkBookingRepository.save(
```
bell-pa, line 60: This would be easier to read if the formatter didn't insist on folding on to multiple lines.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/CourtController.kt
```diff
@@ -56,4 +60,11 @@ class CourtController(private val courtService: CourtService) {
   @ApiOperation(value = "Create a Video Link Booking")
   fun createVideoLinkBooking(@RequestBody @Valid videoLinkBookingSpecification: VideoLinkBookingSpecification) =
     courtService.createVideoLinkBooking(videoLinkBookingSpecification)
+
+  @PostMapping(path = ["/appointment-linker"])
+  @ResponseStatus(HttpStatus.NO_CONTENT)
+  @ApiOperation(value = "Create Video Link Bookings for dangling Video Link Appointments")
+  fun linkDanglingAppointments() {
```
bell-pa, line 67: Should this be guarded by membership of some role?

src/main/resources/db/migration/V14__video_link_booking_unique_constraints.sql
bell-pa: Done.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/VideoLinkAppointmentLinker.kt
```diff
@@ -0,0 +1,118 @@
...
+
+    val appointmentsByEndTime = prisonAppointments
+      .filter { appointmentsByAppointmentId.containsKey(it.eventId) }
+      .associateBy({ it.endTime }, { appointmentsByAppointmentId[it.eventId] })
+      .toMutableMap()
+
+    log.info("bookingId $bookingId: Found ${appointmentsByAppointmentId.size} PRE VideoLinkAppointments, Matched ${appointmentsByEndTime.size} with prison appointments")
+
+    return ({ mainAppointment -> appointmentsByEndTime.remove(mainAppointment?.startTime) })
+  }
```
bell-pa, line 97: Don't want a pre/post appointment to be associated with more than one main appointment so remove when matched.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/CourtController.kt
JamieDraperUK: Feels like it should but then again I suspect this code will eventually be removed once all the appointments have been linked?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/VideoLinkAppointmentLinker.kt
JamieDraperUK: Yeah it would be good to batch the calls as requesting appointments are a bit slow sometimes.

andrewrlee (dismissed):
LGTM

JamieDraperUK (approved):
:+1:

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/AppointmentLinkerApplication.kt
```diff
@@ -0,0 +1,76 @@
+package uk.gov.justice.digital.hmpps.whereabouts
+/*
+This runs, but
+* I can't work out how to get the gradle build to work with two main classes.
+* The configuration pollutes the main app configuration and breaks the tests. (needs moving elsewhere)
```
andrewrlee, line 5: This is all commented out. do we need this or can it go?

andrewrlee (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/repository/AppointmentLinkerSupportTest.kt
mikehalmamoj: Yes I also dislike that rule and considered suggesting we turn it off.  However I like everything else the formatter does and I'm not sure we want to get into the game of arguing over our own ruleset so I decided to suck it up.

PR #224, opened 2020-12-02
DCS-701 create delete end point

louisenorris:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/CourtController.kt
```diff
@@ -67,4 +70,12 @@ class CourtController(
   fun linkDanglingAppointments() {
     appointmentLinker.linkAppointments()
   }
+
+  @DeleteMapping(consumes = [MediaType.APPLICATION_JSON_VALUE], path = ["/video-link-bookings/{videoBookingId}"])
+  @ResponseStatus(HttpStatus.NO_CONTENT)
```
bell-pa, line 75: Being a DELETE operations there isn't any body to consume. There's no need for a 'consumes' attribute.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/CourtService.kt
```diff
@@ -147,4 +148,44 @@ class CourtService(
...
     )
+
+    @Transactional
+    fun deleteVideoLinkBooking(videoBookingId: Long) {
+      val booking = videoLinkBookingRepository.findById(videoBookingId).orElseThrow {
+        EntityNotFoundException("Video link booking with id $videoBookingId not found")
+      }
+
+      booking.toAppointments().forEach { prisonApiService.deleteAppointment(it.appointmentId) }
+      booking.toAppointments().forEach { videoLinkAppointmentRepository.deleteById(it.id!!) }
```
bell-pa, line 159: You don't need to delete the appointments.  JPA cascades should delete the appointments when you delete the booking.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/CourtService.kt
```diff
@@ -147,4 +148,44 @@ class CourtService(
...
+      videoLinkBookingRepository.deleteById(booking.id!!)
+
```
bell-pa, line 161: Or videoLinkBookingRepository.deleteById(videoBookingId)

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/PrisonApiService.java
```diff
@@ -212,5 +215,20 @@ public PrisonAppointment getPrisonAppointment(long appointmentId) {
                 .blockOptional()
                 .orElse(null);
     }
+
+    public void deleteAppointment(final Long appointmentId) {
+        try {
+            webClient.delete()
+                    .uri("/appointments/{appointmentId}", appointmentId)
+                    .retrieve()
+                    .bodyToMono(String.class)
```
bell-pa, line 224: The DELETE response doesn't have a body.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/PrisonApiService.java
```diff
@@ -212,5 +215,20 @@ public PrisonAppointment getPrisonAppointment(long appointmentId) {
...
+                    .block();
+        } catch (WebClientResponseException e) {
+            if (!e.getStatusCode().equals(NOT_FOUND)) {
+                throw e;
```
bell-pa, line 228: Throwing an exception here will stop the delete and yet you could carry on, delete any remaining appointments in prison-api and then delete the VideoLinkBooking in whereabouts-api.

andrewrlee: This is ignoring 404s and only stopping on 500s - that makes sense right? If you can't delete the first appointment in nomis it would be better to error and let the user retry

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/PrisonApiService.java
andrewrlee: this is how all the other void methods are consumed. Not too sure how to handle this otherwise?

andrewrlee: ahh `.toBodilessEntity().block()`

GurnankCheema: updated with   .toBodilessEntity().block()

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/PrisonApiService.java
GurnankCheema: removed try catch. Replaced with onResumeError(...

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/CourtService.kt
GurnankCheema: reved delete operations

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/CourtController.kt
GurnankCheema: consumes = [MediaType.APPLICATION_JSON_VALUE]  removed

bell-pa (approved):
LGTM

PR #226, opened 2020-12-03
NN-3081 add custom telemetry event

JamieDraperUK:


build.gradle.kts
```diff
@@ -18,7 +18,7 @@ dependencies {
```
JamieDraperUK, line 18: Got a ticker on the board to start migrating to the latest dps gradle plugin.

mikehalmamoj: Just to 1.1.2 for now - there are some issues with v2.0.0

mikehalmamoj (approved):

PR #232, opened 2020-12-09
DCS-700 Add an end-point like 'GET /video-link-boookings/date/2020-11-12?court=Wimbledon'

bell-pa:
court query parameter is optional.


andrewrlee (approved):
LGTM

PR #238, opened 2020-12-17
DCS-747 Create an 'Amend video link booking' end-point

bell-pa:
Added some validation for appointment start and end times prior to attempting to create appointments in prison-api.


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/CourtService.kt
```diff
@@ -236,6 +241,27 @@ class CourtService(
...
+    specification.post?.let { validateVideoLinkAppointment(it, "Post") }
+  }
+
+  private fun validateVideoLinkBookingUpdateSpecification(specification: VideoLinkBookingUpdateSpecification) {
+    validateVideoLinkAppointment(specification.main, "Main")
+    specification.pre?.let { validateVideoLinkAppointment(it, "Pre") }
+    specification.post?.let { validateVideoLinkAppointment(it, "Post") }
+  }
+
+  private fun validateVideoLinkAppointment(appointment: VideoLinkAppointmentSpecification, prefix: String) {
```
mikehalmamoj, line 256: This reads well and I can follow easily so it's fine.  But I was wondering how it would look if you were to make this an extension function called `VideoLinkAppointmentSpecification.validate` - you might lose a lot of the verbosity?

mikehalmamoj (dismissed):

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/CourtService.kt
bell-pa: Excellent suggestion. I'll get on with it.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/CourtService.kt
```diff
@@ -236,6 +242,33 @@ class CourtService(
...
+    if (appointment.startTime.isBefore(LocalDateTime.now(clock))) {
+      throw ValidationException("$prefix appointment start time must be in the future.")
+    }
+    if (!appointment.startTime.isBefore(appointment.endTime)) {
+      throw ValidationException("$prefix appointment start time must precede end time.")
+    }
+    appointment.locationId?.let {
+      val location = prisonApiService.getLocation(it) ?: throw ValidationException("$prefix locationId $it not found in NOMIS.")
+      if (VIDEO_LINK_APPOINTMENT_LOCATION_TYPE != location.locationType) {
```
andrewrlee, line 266: can we do this? the prison side doesn't limit the location list to just video link appointments

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/CourtService.kt
```diff
@@ -236,6 +242,33 @@ class CourtService(
...
+
+  private fun validateVideoLinkAppointment(appointment: VideoLinkAppointmentSpecification, prefix: String) {
+    if (appointment.startTime.isBefore(LocalDateTime.now(clock))) {
+      throw ValidationException("$prefix appointment start time must be in the future.")
+    }
+    if (!appointment.startTime.isBefore(appointment.endTime)) {
+      throw ValidationException("$prefix appointment start time must precede end time.")
+    }
+    appointment.locationId?.let {
+      val location = prisonApiService.getLocation(it) ?: throw ValidationException("$prefix locationId $it not found in NOMIS.")
```
andrewrlee, line 265: meh... would be slightly more verbose but could potentially get locations for an agency - would be a single call rather than upto 3 and you could also check all locations existed in the same prison? 🤷

bell-pa: 6 vs 12/2?  The individual locationDtos have an 'agencyId' field in the response so I could ensure that they all had the same agencyId...  That's sort of contrary to removing the location type check above though.  Should I put business logic in whereabouts-api (beyond safety checks) or leave that to BVL and prisonstaffhub?

bell-pa: There are 85 locations of type 'APP' at WWI. 16 of those have locationType == 'VIDE'

bell-pa: To look up locations by agencyId I need the agencyId.  I can either hard-wire that to 'WWI' or ask prison-api for offender details for the bookingId and take the offender's agencyId - or the client has to send the agencyId in the create or update booking request.

bell-pa: Decided to stay with individual look-ups for now.  Can change if you think it is important though...

andrewrlee (approved):
LGTM

PR #241, opened 2020-12-22
DCS-761 Adding endpoint to allow retrieving bookings for a prison. 

andrewrlee:
The existing endpoint only returns bookings for Wandsworth


petergphillips (approved):
LGTM

PR #246, opened 2021-01-07
DCS-774 Fixing issue around nullability and end dates

andrewrlee:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/CourtService.kt
```diff
@@ -246,14 +252,20 @@ class CourtService(
...
+    scheduledAppointment?.takeIf { hasAnEndDate(it) }?.let {
       VideoLinkBookingResponse.LocationTimeslot(
         locationId = it.locationId,
         startTime = it.startTime,
-        endTime = it.endTime
+        endTime = it.endTime!!
       )
     }
 
+  private fun hasAnEndDate(it: ScheduledAppointmentDto): Boolean {
```
bell-pa, line 264: Could be an extension function. Don't need it though.

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/services/CourtServiceTest.kt
```diff
@@ -1047,6 +1047,74 @@ class CourtServiceTest {
...
+            VideoLinkBooking(
+              id = 1000L,
+              pre = VideoLinkAppointment(
+                id = 10001,
+                bookingId = 1000,
+                appointmentId = 1001,
+                court = "Wimbledon",
+                hearingType = HearingType.PRE
+              ),
+              main = VideoLinkAppointment(
```
bell-pa, line 1094: What happens if the main appointment isn't valid? 
Remove the booking from the result list?

bell-pa (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/services/CourtServiceTest.kt
andrewrlee: fixed that

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/CourtService.kt
andrewrlee: ooh that would be nice

PR #250, opened 2021-01-15
DCS-778 Fix swagger documentation

bell-pa:
* LocalDate, LocalTime, LocalDate etcTime and ZonedDateTime now mapped to String instead of java.util.Date
* Fixed swagger doc for Location  by re-naming to LocationIdAndDescription because SpringFox couldn't handle  two classes having the same name.


GurnankCheema (approved):
LGTM

PR #252, opened 2021-01-19
DCS-794 Fix location description in new endpoint

andrewrlee:
Description is all uppercase, general pattern is to prefer user description if available of fall back to description


petergphillips (approved):
LGTM

PR #254, opened 2021-01-27
DCS-811 removed madeInCourt flag when updating booking

GurnankCheema:
The madeInCourt field shouldn't be changed after creation


bell-pa (approved):
LGTM

PR #263, opened 2021-03-08
DCS-899 :bug: Availability check not taking other appointments into account. 

bell-pa:
Stopped filtering appointments by appointmentTypeCode == 'VLB'.


petergphillips (approved):
LGTM

PR #267, opened 2021-03-24
DCS-888 Book a Video Link - Analytics. 

bell-pa:
* Simplify transaction boundary handling
* Fix bug in EventStoreListener and its test class


src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/CourtIntegrationTest.kt
```diff
@@ -32,10 +34,27 @@ class CourtIntegrationTest : IntegrationTest() {
   @Autowired
   lateinit var objectMapper: ObjectMapper
 
+  @Autowired
+  lateinit var telemetryClient: TelemetryClient
```
andrewrlee, line 38: would it not be better to mock this out?

bell-pa: Yes. Done.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/court/CourtService.kt
```diff
@@ -33,11 +33,9 @@ class CourtService(
   private val videoLinkAppointmentRepository: VideoLinkAppointmentRepository,
   private val videoLinkBookingRepository: VideoLinkBookingRepository,
   private val clock: Clock,
-  private val eventStoreListener: VideoLinkBookingEventListener,
-  private val applicationInsightsEventListener: VideoLinkBookingEventListener,
+  private val videoLinkBookingEventListener: VideoLinkBookingEventListener,
```
sp-watson, line 36: it doesn't matter too much, but I preferred keeping the EventStoreListener and AppInsightsListener as separate dependencies, as that guarantees the EventStore is updated before the AppInsights message is triggered

sp-watson (approved):

PR #290, opened 2021-05-17
NN-3374: Return extra data from appointments search endpoint

sp-watson:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AppointmentService.kt
```diff
@@ -74,14 +75,19 @@ class AppointmentService(
   }
 }
 
-private fun makeAppointmentDto(scheduledAppointmentDto: ScheduledAppointmentDto): AppointmentDto = AppointmentDto(
+private fun makeAppointmentDto(scheduledAppointmentDto: ScheduledAppointmentSearchDto): AppointmentSearchDto = AppointmentSearchDto(
```
JamieDraperUK, line 78: Might need to tweak the method name

JamieDraperUK (approved):
LGTM

PR #292, opened 2021-05-18
DCS-703 (almost) normalise video_link_booking and video_link_appointment

bell-pa:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/CourtController.kt
```diff
@@ -103,9 +103,7 @@ class CourtController(
     )
     @RequestParam(name = "court", required = false)
     court: String?
-  ): List<VideoLinkBookingResponse> {
-    return courtService.getVideoLinkBookingsForPrisonAndDateAndCourt("WWI", date, court)
-  }
+  ) = courtService.getVideoLinkBookingsForPrisonAndDateAndCourt("WWI", date, court)
```
andrewrlee, line 106: reckon we can probably get rid of this endpoint now.. not as part of this PR but we could check to see if its still being used and get rid of it

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/dto/AppointmentDtos.kt
```diff
@@ -117,3 +118,38 @@ data class AppointmentCreatedDto(
   val mainAppointmentId: Long,
   val recurringAppointments: Set<Long>? = null
 )
+
+fun VideoLinkBooking.preAppointmentDto() =
+  if (pre != null)
```
andrewrlee, line 123: meh... guess  you could use `.let` (or one of the other fancy methods ) to get rid of the if/else and the `!!`

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/model/VideoLinkAppointment.kt
```diff
@@ -23,14 +20,6 @@ enum class HearingType {
...
-  val appointmentId: Long,
-  val court: String,
 
-  @Enumerated(EnumType.STRING)
-  val hearingType: HearingType,
-
-  @CreatedBy
-  var createdByUsername: String? = null,
-  val madeByTheCourt: Boolean? = true
+  val appointmentId: Long
```
andrewrlee, line 24: cool so leaving so can potentially add start/end times etc.. if in the future this becomes the master of VLB data

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AppointmentService.kt
```diff
@@ -125,19 +125,7 @@ private fun makeAppointmentDto(offenderNo: String, prisonAppointment: PrisonAppo
...
-    bookingId = videoLinkAppointment.bookingId,
-    appointmentId = videoLinkAppointment.appointmentId,
-    court = videoLinkAppointment.court,
-    hearingType = videoLinkAppointment.hearingType,
-    createdByUsername = videoLinkAppointment.createdByUsername,
-    madeByTheCourt = videoLinkAppointment.madeByTheCourt
-
+    main = videoLinkBooking.mainAppointmentDto(),
+    pre = videoLinkBooking.preAppointmentDto(),
+    post = videoLinkBooking.postAppointmentDto()
```
andrewrlee, line 130: nice!

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/court/CourtService.kt
```diff
@@ -46,22 +46,18 @@ class CourtService(
...
-          bookingId = it.bookingId,
-          appointmentId = it.appointmentId,
-          hearingType = it.hearingType,
-          court = it.court,
-          createdByUsername = it.createdByUsername,
-          madeByTheCourt = it.madeByTheCourt
-        )
-      }.toSet()
-  }
+        with(it) {
```
andrewrlee, line 52: finding this a bit difficult to read - it reads like only a single appointment can be retrieved from a booking?

andrewrlee: how is only one appointment returned for a booking?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/court/CourtService.kt
```diff
@@ -102,43 +101,16 @@ class CourtService(
...
-        bookingId = bookingId,
-        appointmentId = it.eventId,
-        court = court,
-        hearingType = HearingType.POST,
-        madeByTheCourt = madeByTheCourt
-      )
-    }
+    booking.main = VideoLinkAppointment(appointmentId = mainEvent.eventId)
+    booking.pre = preEvent?.let { VideoLinkAppointment(appointmentId = it.eventId) }
+    booking.post = postEvent?.let { VideoLinkAppointment(appointmentId = it.eventId) }
```
andrewrlee, line 113: nice!

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/locationfinder/AppointmentLocationsService.kt
```diff
@@ -32,7 +32,15 @@ class AppointmentLocationsService(
...
         .findAllById(vlbIdsToExclude)
-        .flatMap { booking -> booking.toAppointments().map { it.appointmentId } }
+        .asSequence()
+        .flatMap {
+          sequenceOf(
+            it.main.appointmentId,
+            it.post?.appointmentId,
+            it.pre?.appointmentId
+          )
+            .filterNotNull()
```
andrewrlee, line 42: meh.. could add a `toAppointmentIds`method to VLB, like the current `toAppointmentDtos`

that uses `listOfNotNull` presumably there is a `setOfNotNull`  too.

src/main/resources/db/migration/V18__video_link_booking_normalisation.sql
```diff
@@ -0,0 +1,41 @@
+create table if not exists video_link_appointment_pre_DCS_703 as table video_link_appointment;
+create table if not exists video_link_booking_pre_DCS_703 as table video_link_booking;
+
+alter table video_link_booking
+    add column booking_id bigint;
```
andrewrlee, line 5: need an index for this?

src/test/resources/uk/gov/justice/digital/hmpps/whereabouts/integration/courtAppointments.json
```diff
@@ -5,7 +5,7 @@
       "bookingId": 1,
       "appointmentId": 1,
       "court": "York",
-      "hearingType": "PRE"
+      "hearingType": "MAIN"
```
andrewrlee, line 8: Why did this change?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/court/CourtService.kt
bell-pa: This is really a legacy method used by PSH. Previously it searched for VideoLinkAppointments using VideoLinkAppointmentRepository. Now court, madeByTheCourt have been moved to the parent VideoLinkBooking. The repository query returns any booking where any appointmentid matches supplied appointmentIds set.  Then have to work out which one(s) matched and return DTOs for each.

On reflection I think I could have done some sort of join from appointment to booking. to return an appointment with its booking, or , event better, constructed the DTO in the query. (I think I might try that).

bell-pa: Reviewing this code myself I realise it has a bug.
Originally this method used VideoLinkAppointmentRepository to retrieve an appointment with it's bookingId, court, madeByCourt.  
Those properties have been moved to the parent VideoLinkBooking so the original query is no longer sufficient.  Each appointment must be retrieved with its parent booking (or at least the properties of the parent booking). Since there are three separate releationships between a booking and an appointment that is a bit ugly. 
If we were free to redesign the tables we might possibly have a composite PK for appointment of videoLinkBookingId + HearingType (perhaps). But we have to retain the surrogate key of VideoLinkAppointment because it is used by PSH.
The bug arises because the same VLB can be returned for up to 3 different VLA - where the relationship(s) are some combination of pre, main and post.  DTOs should be constructed and returned for all matching relationships, but as it stands the code will return at most one.

src/test/resources/uk/gov/justice/digital/hmpps/whereabouts/integration/courtAppointments.json
bell-pa: Every video link booking must have a MAIN appointment. You can't have a solitary PRE appointment. I've move this JSON in-line becuase its the only place it is used.

PR #294, opened 2021-05-20
NN-3375:

JamieDraperUK:
- Create a new data model to store information around recurring appointments so that they can be deleted as a whole.


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/dto/AppointmentDtos.kt
```diff
@@ -20,16 +21,68 @@ data class CreateBookingAppointment(
...
+    required = false,
+    value = "The estimated date time the appointment will end",
+    example = "2021-05-23T17:00:00"
+  )
+  val endTime: LocalDateTime? = null,
+  @ApiModelProperty(required = false, value = "Describes how many times this appointment is to be repeated")
+  val repeat: Repeat? = null
+)
+
+data class CreatePrisonAppointment(
```
sp-watson, line 54: Maybe this is for later refactoring, but this data structure (and I suspect others in this file) are not DTOs.. and this class is getting too big IMO... so we ought to put them in a class (prob in the same package as the service)

sp-watson (approved):

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/dto/AppointmentDtos.kt
JamieDraperUK: Agreed

PR #299, opened 2021-05-24
NN-3407:

JamieDraperUK:
- Create new endpoint to delete recurring appointments
- Refactor data model


src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/controllers/AppointmentControllerTest.kt
```diff
@@ -175,6 +176,22 @@ class AppointmentControllerTest : TestController() {
     }
   }
 
+  @Nested
```
JamieDraperUK, line 179: More tests coming in the next PR

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AppointmentService.kt
```diff
@@ -101,29 +110,63 @@ class AppointmentService(
...
 
+  @Transactional
+  fun deleteAppointment(appointmentId: Long) {
+    prisonApiService.getPrisonAppointment(appointmentId)
+      ?: throw EntityNotFoundException("Appointment $appointmentId does not exist")
+
+    val videoLinkBooking: VideoLinkBooking? =
+      videoLinkBookingRepository.findByMainAppointmentIds(listOf(appointmentId)).firstOrNull()
+
+    if (videoLinkBooking != null) {
```
sp-watson, line 134: Just checking - we can't have recurring VLBs?

sp-watson (approved):

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AppointmentService.kt
JamieDraperUK: We can't at the minute. The design team are confident that we won't in the future.

PR #302, opened 2021-05-25
NN-3407:

JamieDraperUK:
- Fix regression in returning a recurring appointment

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AppointmentService.kt
+++ b/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AppointmentService.kt
@@ -78,18 +78,19 @@ class AppointmentService(
     val mainAppointmentDetails: PrisonAppointment = prisonApiService.getPrisonAppointment(appointmentId)
       ?: throw EntityNotFoundException("Appointment $appointmentId does not exist")
 
-    val offenderNo = prisonApiService.getOffenderNoFromBookingId(mainAppointmentDetails.bookingId)
+    val offenderNo = try {
+      prisonApiService.getOffenderNoFromBookingId(mainAppointmentDetails.bookingId)
+    } catch (e: Exception) {
+      null
+    }
 
     val videoLinkBooking: VideoLinkBooking? =
       videoLinkBookingRepository.findByMainAppointmentIds(listOf(appointmentId)).firstOrNull()
 
     val recurringAppointment: RecurringAppointment? =
       recurringAppointmentRepository.findRecurringAppointmentByRelatedAppointmentsContains(
-        RelatedAppointment(
-          appointmentId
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AppointmentService.kt
```diff
@@ -78,18 +78,19 @@ class AppointmentService(
     val mainAppointmentDetails: PrisonAppointment = prisonApiService.getPrisonAppointment(appointmentId)
       ?: throw EntityNotFoundException("Appointment $appointmentId does not exist")
 
-    val offenderNo = prisonApiService.getOffenderNoFromBookingId(mainAppointmentDetails.bookingId)
+    val offenderNo = try {
+      prisonApiService.getOffenderNoFromBookingId(mainAppointmentDetails.bookingId)
+    } catch (e: Exception) {
+      null
```
sp-watson, line 84: We need a good reason to ignore exceptions - is it the case here that this is just extra data that shouldn't stop the rest of the data from being returned?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AppointmentService.kt
```diff
@@ -109,8 +110,14 @@ class AppointmentService(
...
     createAppointmentSpecification.repeat?.let {
+
+      val recurringAppointmentIds: Set<Long> = appointmentCreated.recurringAppointmentEventIds?.toSet() ?: emptySet()
+
+      val appointmentIds =
+        setOf(appointmentCreated.appointmentEventId).union(recurringAppointmentIds)
+
       val recurringAppointment =
-        makeMainRecurringAppointment(appointmentCreated, createAppointmentSpecification.repeat)
+        makeRecurringAppointment(appointmentIds = appointmentIds, repeat = createAppointmentSpecification.repeat)
```
sp-watson, line 120: That first "=" seems superfluous - does Kotlin need that?

sp-watson (approved):

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AppointmentService.kt
JamieDraperUK: I did some testing and for some offenders I was getting a 404. I'll do some more testing and maybe remove or add some logging.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AppointmentService.kt
JamieDraperUK: Good point will update in the next PR

PR #303, opened 2021-05-25
NN-3407:

JamieDraperUK:
- Updates to use the new create appointment prison api data contract

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/AppointmentController.kt
+++ b/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/AppointmentController.kt
@@ -67,7 +67,7 @@ class AppointmentController(private val appointmentService: AppointmentService) 
       )
     ]
   )
-  fun createAppointment(@RequestBody createAppointmentSpecification: CreateAppointmentSpecification): CreatedAppointmentDetailsDto =
+  fun createAppointment(@RequestBody createAppointmentSpecification: CreateAppointmentSpecification): List<CreatedAppointmentDetailsDto> =
     appointmentService.createAppointment(createAppointmentSpecification)
 
   @DeleteMapping(path = ["/{id}"])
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/AppointmentController.kt
```diff
@@ -67,7 +67,7 @@ class AppointmentController(private val appointmentService: AppointmentService)
       )
     ]
   )
-  fun createAppointment(@RequestBody createAppointmentSpecification: CreateAppointmentSpecification): CreatedAppointmentDetailsDto =
+  fun createAppointment(@RequestBody createAppointmentSpecification: CreateAppointmentSpecification): List<CreatedAppointmentDetailsDto> =
```
sp-watson, line 70: Hmm.. didn't really notice in last PR, but do we need to return this data? I suppose we have to be consistent with prison-api?!

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AppointmentService.kt
```diff
@@ -105,16 +105,12 @@ class AppointmentService(
   }
 
   @Transactional
-  fun createAppointment(createAppointmentSpecification: CreateAppointmentSpecification): CreatedAppointmentDetailsDto {
+  fun createAppointment(createAppointmentSpecification: CreateAppointmentSpecification): List<CreatedAppointmentDetailsDto> {
     val appointmentCreated =
```
sp-watson, line 109: Small thing - should be plural

sp-watson (approved):

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/AppointmentController.kt
JamieDraperUK: We don't need to really but felt like the correct thing to do.

PR #305, opened 2021-05-25
DCS-1002 Derive court name from court id where possible.

bell-pa:
CourtService renamed VideoLinkBookingService because it handles VideoLinkBookings and VideoLinkAppointments.
New CourtService: information about courts - court identifier, court name, look-ups etc.
CourtController renamed VideoLinkBookingController.


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/model/Court.kt
```diff
@@ -9,11 +9,11 @@ import javax.persistence.Table
 @Entity
 @Table(name = "enabled_court")
 @ApiModel(description = "Video Link Booking related information for a court")
-data class Court(
```
bell-pa, line 12: Don't use data class for entities and don't use immutable properties they don't work well with Hibernate.
In this case equals and hash code don't need overriding because these objects aren't used in sets.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AppointmentService.kt
```diff
@@ -26,17 +26,19 @@ import uk.gov.justice.digital.hmpps.whereabouts.model.VideoLinkBooking
...
 import javax.persistence.EntityNotFoundException
 import javax.transaction.Transactional
 
 @Service
 class AppointmentService(
+  private val courtService: CourtService,
   private val prisonApiService: PrisonApiService,
   private val videoLinkBookingRepository: VideoLinkBookingRepository,
   private val recurringAppointmentRepository: RecurringAppointmentRepository,
-  private val courtService: CourtService,
```
bell-pa, line 39: CourtService used to look up court name from court id.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AppointmentService.kt
```diff
@@ -207,6 +209,58 @@ class AppointmentService(
     )
   }
 
+  private fun makeVideoLinkBookingAppointmentDto(
```
bell-pa, line 212: Moved inside class so that makeVideoLinkAppointmentDto can access courtService

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/court/CourtService.kt
```diff
@@ -1,321 +1,30 @@
 package uk.gov.justice.digital.hmpps.whereabouts.services.court
```
bell-pa, line 1: Old CourtService renamed VideoLinkBookingService.  
New CourtService here has nothing in common with old CourtService.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/court/CourtService.kt
```diff
@@ -1,321 +1,30 @@
...
-  companion object {
-    val log: Logger = LoggerFactory.getLogger(this::class.java)
+class CourtService(private val courtRepository: CourtRepository) {
+  val courtNames: List<String> by lazy {
+    courts.map { it.name }
   }
 
-  @Transactional(readOnly = true)
-  fun getCourtLocations() = courtRepository.findAll().map { it.name }.toSet()
+  val courts: List<Court> by lazy { courtRepository.findAll(Sort.by("name")) }
```
bell-pa, line 17: Only fetch court data once. It doesn't change.  Additions and revisions to the underlying table enabled_court is likely to be performed by flyway scripts at start up.  All pods should be updated, the first will lock and update tables, the other pods will see the new data.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/court/VideoLinkBookingService.kt
```diff
@@ -0,0 +1,313 @@
+package uk.gov.justice.digital.hmpps.whereabouts.services.court
```
bell-pa, line 1: Renamed from CourtService.
Added CourtService as a dependency.
Only changes are to add lookups for court name -> court Id and court id -> court name when creating bookings and when retrieving bookings for client queries.

src/main/resources/db/migration/V23__revise_court_ids.sql
```diff
@@ -0,0 +1,11 @@
+delete from enabled_court where id = 'DRBYMC';
+
+update enabled_court set id = 'DRBYMC' where id = 'DRBYJC';
+
```
bell-pa, line 4: All court_id will be null before this update.
Some will not be null afterwards.  
On preprod data this catches all but 9 of the video_link_appointment where made_by_court is true. (5 bookings).

andrewrlee (approved):
LGTM

PR #310, opened 2021-05-28
Nn 3422

JamieDraperUK:
- Use new delete multiple appointments endpoint


bell-pa (commented):
Looks OK.
Personally I'd prefer to see a full stack integration test that worked against the database - I think it helps when Hibernate and transactions are involved.  I know that can be tricky though.
Bit concerned about the divergence between PSH and BVL in relation to video link bookings - but that's a separate discussion!

bell-pa (approved):
LGTM

PR #315, opened 2021-06-08
NN-3437: Add first appt start time to recurring appointment info

sp-watson:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/model/RelatedAppointment.kt
```diff
@@ -24,6 +25,7 @@ data class RecurringAppointment(
 
   val repeatPeriod: RepeatPeriod,
   val count: Long,
+  val startTime: LocalDateTime,
```
sp-watson, line 28: Not sure about this name, but given the context is RecurringAppointment maybe it's enough?

JamieDraperUK (approved):
LGTM

PR #317, opened 2021-06-09
NN-3442: Allow for deletion of single appointment in recurring list

sp-watson:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/AppointmentController.kt
```diff
@@ -89,5 +91,8 @@ class AppointmentController(private val appointmentService: AppointmentService)
       )
     ]
   )
-  fun deleteAppointment(@PathVariable(value = "id") id: Long) = appointmentService.deleteAppointment(id)
+  fun deleteAppointment(
+    @PathVariable(value = "id") id: Long,
+    @ApiParam(value = "Whether to delete the whole sequence of recurring appointments to which this appointment belongs.") @RequestParam(value = "deleteRelatedAppointments", required = false, defaultValue = "true") deleteRelatedAppointments: Boolean
+  ) = appointmentService.deleteAppointment(id, deleteRelatedAppointments)
```
sp-watson, line 97: i think we ought to have a separate endpoint to delete all recurring appointments, as we could then use the recurring appointment id which would prevent any nasties around a single appointment being deleted. I was going to do that next

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/model/RelatedAppointment.kt
```diff
@@ -29,7 +29,7 @@ data class RecurringAppointment(
 
   @OneToMany(cascade = [CascadeType.ALL], orphanRemoval = true)
   @JoinColumn(name = "RECURRING_APPOINTMENT_ID")
-  val relatedAppointments: List<RelatedAppointment>? = null
+  val relatedAppointments: MutableList<RelatedAppointment>? = null
```
sp-watson, line 32: Without using a mutableList `orphanRemoval = true` isn't of any use AFAIAA

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/AppointmentController.kt
JamieDraperUK: Good shout

JamieDraperUK (approved):
:+1:

PR #326, opened 2021-06-22
DCS-1042 End point for finding Video Link Booking availability.

bell-pa:
Have a look at story DCS-1042, then look at `VideoLinkBookingOptionsFinderTest`. That class tests the most of business logic for this story and is fairly easy to understand.  
The overall flow is: 

- `VideoLinkBookingController.findAvaliableVideoLinkBookingOptions` ->
- `AppointmentLocationsService.findVideoLinkBookingOptions()` ->
- `VideoLinkBookingOptionsService.findVideoLinkBookingOptions`

`VideoLinkBookingOptionsService.findVideoLinkBookingOptions()` does the interactions with prison-api and the database to retrieve and massage the data that `VideoLinkBookingOptionsFinder` uses to produce its result.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/VideoLinkBookingController.kt
+++ b/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/VideoLinkBookingController.kt
@@ -29,7 +29,10 @@ import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.Appointm
 import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.AppointmentLocationsSpecification
 import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.AvailableLocations
 import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.AvailableVideoLinkBookingLocations
+import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.LocationIdAndDescription
 import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.VideoLinkBookingLocationsSpecification
+import uk.gov.justice.digital.hmpps.whereabouts.services.vlboptionsfinder.VideoLinkBookingOptions
+import uk.gov.justice.digital.hmpps.whereabouts.services.vlboptionsfinder.VideoLinkBookingSearchSpecification
 import java.time.LocalDate
 import javax.validation.Valid
 import javax.validation.constraints.NotNull
@@ -231,4 +234,24 @@ class VideoLinkBookingController(
       specification.preInterval != null
     )
   }
+
+  @GetMapping(
+    path = ["/vlb-locations/{agencyId}"],
+    produces = [MediaType.APPLICATION_JSON_VALUE]
+  )
+  @ResponseStatus(HttpStatus.OK)
+  @ApiOperation(value = "Provide the set of locations within a prison used for Video Link Bookings")
+  fun getVideoLinkBookingLocationsForAgency(
+    @ApiParam(value = "Agency id", required = true)
+    @PathVariable("agencyId")
+    agencyId: String
+  ): List<LocationIdAndDescription> = appointmentLocationsService.allVideoLinkLocationsForAgency(agencyId)
+
+  @PostMapping()
+  fun findAvaliableVideoLinkBookingOptions(
+    @Valid
+    @RequestBody
+    specification: VideoLinkBookingSearchSpecification
+  ): VideoLinkBookingOptions =
+    appointmentLocationsService.finVideoLinkBookingOptions(specification)
 }
--- a/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/VideoLinkBookingController.kt
+++ b/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/VideoLinkBookingController.kt
@@ -253,5 +253,5 @@ class VideoLinkBookingController(
     @RequestBody
     specification: VideoLinkBookingSearchSpecification
   ): VideoLinkBookingOptions =
-    appointmentLocationsService.finVideoLinkBookingOptions(specification)
+    appointmentLocationsService.findVideoLinkBookingOptions(specification)
 }
--- a/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/VideoLinkBookingController.kt
+++ b/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/VideoLinkBookingController.kt
@@ -29,7 +29,6 @@ import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.Appointm
 import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.AppointmentLocationsSpecification
 import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.AvailableLocations
 import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.AvailableVideoLinkBookingLocations
-import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.LocationIdAndDescription
 import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.VideoLinkBookingLocationsSpecification
 import uk.gov.justice.digital.hmpps.whereabouts.services.vlboptionsfinder.VideoLinkBookingOptions
 import uk.gov.justice.digital.hmpps.whereabouts.services.vlboptionsfinder.VideoLinkBookingSearchSpecification
@@ -234,18 +233,6 @@ class VideoLinkBookingController(
       specification.preInterval != null
     )
   }
-
-  @GetMapping(
-    path = ["/vlb-locations/{agencyId}"],
-    produces = [MediaType.APPLICATION_JSON_VALUE]
-  )
-  @ResponseStatus(HttpStatus.OK)
-  @ApiOperation(value = "Provide the set of locations within a prison used for Video Link Bookings")
-  fun getVideoLinkBookingLocationsForAgency(
-    @ApiParam(value = "Agency id", required = true)
-    @PathVariable("agencyId")
-    agencyId: String
-  ): List<LocationIdAndDescription> = appointmentLocationsService.allVideoLinkLocationsForAgency(agencyId)
 
   @PostMapping()
   fun findAvaliableVideoLinkBookingOptions(
--- a/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/VideoLinkBookingController.kt
+++ b/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/VideoLinkBookingController.kt
@@ -30,6 +30,7 @@ import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.Appointm
 import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.AvailableLocations
 import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.AvailableVideoLinkBookingLocations
 import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.VideoLinkBookingLocationsSpecification
+import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.VideoLinkBookingOptionsService
 import uk.gov.justice.digital.hmpps.whereabouts.services.vlboptionsfinder.VideoLinkBookingOptions
 import uk.gov.justice.digital.hmpps.whereabouts.services.vlboptionsfinder.VideoLinkBookingSearchSpecification
 import java.time.LocalDate
@@ -42,7 +43,8 @@ import javax.validation.constraints.NotNull
 class VideoLinkBookingController(
   private val courtService: CourtService,
   private val videoLinkBookingService: VideoLinkBookingService,
-  private val appointmentLocationsService: AppointmentLocationsService
+  private val appointmentLocationsService: AppointmentLocationsService,
+  private val videoLinkBookingOptionsService: VideoLinkBookingOptionsService,
 ) {
   @GetMapping(produces = [MediaType.APPLICATION_JSON_VALUE], path = ["/all-courts"])
   @ResponseStatus(HttpStatus.OK)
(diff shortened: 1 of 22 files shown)


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/VideoLinkBookingController.kt
```diff
@@ -231,4 +233,16 @@ class VideoLinkBookingController(
       specification.preInterval != null
     )
   }
+
+  @PostMapping(
+    path = ["/video-link-booking-check"],
+    consumes = [MediaType.APPLICATION_JSON_VALUE],
+    produces = [MediaType.APPLICATION_JSON_VALUE]
+  )
+  fun findAvaliableVideoLinkBookingOptions(
```
mikehalmamoj, line 242: Going from the convention in this class, does this also need an @ApiOperation?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/VideoLinkBookingController.kt
```diff
@@ -231,4 +233,16 @@ class VideoLinkBookingController(
...
+    @Valid
+    @RequestBody
+    specification: VideoLinkBookingSearchSpecification
```
mikehalmamoj, line 245: I guess the only consumer is your UI hence the lack of OpenApi docs around the request bodies?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/locationfinder/AppointmentLocationsService.kt
```diff
@@ -46,6 +53,26 @@ class AppointmentLocationsService(
       .getScheduledAppointments(specification.agencyId, specification.date)
       .filter { it.endTime != null }
 
+  fun findVideoLinkBookingOptions(specification: VideoLinkBookingSearchSpecification): VideoLinkBookingOptions {
+    val excludedAppointmentIds =
+      specification.vlbIdToExclude?.let { appointmentIdsFromBookingIds(listOf(it)) } ?: emptyList()
+
+    val locationIds = setOfNotNull(
```
mikehalmamoj, line 60: :smile:

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/locationfinder/AppointmentLocationsService.kt
```diff
@@ -46,6 +53,26 @@ class AppointmentLocationsService(
...
+      specification.preAppointment?.locationId,
+      specification.mainAppointment.locationId,
+      specification.postAppointment?.locationId
+    )
+
+    val scheduledAppointments =
+      prisonApiService
+        .getScheduledAppointments(specification.agencyId, specification.date).asSequence()
```
mikehalmamoj, line 68: Just wondering what the rationale was for using Sequences here?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/vlboptionsfinder/OptionsGenerator.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.format.annotation.DateTimeFormat
+import org.springframework.stereotype.Component
+import java.time.Duration
+import java.time.LocalTime
+
+@Component
+class OptionsGenerator(
+  @DateTimeFormat(iso = DateTimeFormat.ISO.TIME)
+  @Value("\${video-link-booking.day-start}")
```
mikehalmamoj, line 12: I quite like using `@ConfigurationProperties` at the moment.  If you load them into a bean then you can make it a `@SpyBean` in integration tests meaning you don't need to start a new Spring context when testing different values in the properties.  Or maybe we just write too many integration tests on our projects :smiley:

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/vlboptionsfinder/OptionsGenerator.kt
```diff
@@ -0,0 +1,33 @@
...
+
+  @Value("\${video-link-booking.step}")
+  val step: Duration
+) {
+  fun getOptionsInPreferredOrder(preferredOption: VideoLinkBookingOption): Sequence<VideoLinkBookingOption> =
+    durations(step)
+      .map { dayStart.plus(it) }
+      .map { preferredOption.copyStartingAt(it) }
+      .takeWhile { it.endsOnOrBefore(dayEnd) }
+      .sortedBy { Duration.between(preferredOption.main.interval.start, it.main.interval.start).abs() }
```
mikehalmamoj, line 27: nice

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/vlboptionsfinder/VideoLinkBookingOptionsFinder.kt
```diff
@@ -0,0 +1,158 @@
...
+
+    // The time when the room last became unoccupied.
+    // This will be null when the room is currently occupied.
+    var freePeriodStarted: LocalTime? = LocalTime.MIN
+
+    events
+      .plus(StartEvent(LocalTime.MAX))
+      .sortedBy { it.time }
+      .forEach {
+        if (it.time > previousEventTime) {
```
mikehalmamoj, line 43: This is clearly quite complicated hence much more comments than code but it feels like you could change this to a couple of `when` cases and also move the comments to explain each?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/locationfinder/AppointmentLocationsService.kt
```diff
@@ -36,8 +40,11 @@ class AppointmentLocationsService(
         .toSet()
 
   private fun fetchVideoLinkBookingLocations(specification: AppointmentLocationsSpecification) =
+    allVideoLinkLocationsForAgency(specification.agencyId)
+
+  fun allVideoLinkLocationsForAgency(agencyId: String): List<LocationIdAndDescription> =
```
sp-watson, line 45: Can this be private?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/vlboptionsfinder/VideoLinkBookingOptionsFinder.kt
```diff
@@ -0,0 +1,158 @@
...
+  fun emptyPeriods() = emptyPeriods.navigableKeySet().map {
+    Pair(it, emptyPeriods[it])
+  }
+
+  /**
+   * Answer the question 'is this Timeline free of bookings during the specified interval'.
+   */
+  fun isFreeForInterval(interval: Interval): Boolean {
+    if (interval.end.isBefore(interval.start)) {
+      throw IllegalArgumentException("start must precede end")
```
mikehalmamoj, line 97: Would this be better pushed up to `Interval`?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/locationfinder/AppointmentLocationsService.kt
bell-pa: To avoid creating the intermediate collections that would arise from using a list or similar.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/vlboptionsfinder/VideoLinkBookingOptionsFinder.kt
bell-pa: Interesting.  I'll try that out.

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/CourtIntegrationTest.kt
```diff
@@ -524,4 +524,41 @@ class CourtIntegrationTest(
       )
     }
   }
+
+  @Nested
+  inner class VideoLinkBookingCheck {
+    @Test
+    fun happyFlow() {
```
mikehalmamoj, line 531: Maybe have 1 negative test too?

src/main/resources/application.yml
```diff
@@ -84,3 +84,9 @@ management:
 sqs:
   provider: aws
   endpoint.region: eu-west-2
+
+video-link-booking:
+  day-start: "09:00"
+  day-end: "16:00"
+  step: "PT15M"
+  max-alternatives: 3
```
sp-watson, line 92: I like the use of config, but maybe worth putting these options in a sub-group that indicates they are only used when checking availability or finding options? As it stands, it suggests to me that all VLBs are between these times?!

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/services/locationfinder/AppointmentLocationServiceTest2.kt
```diff
@@ -0,0 +1,300 @@
...
+import uk.gov.justice.digital.hmpps.whereabouts.services.vlboptionsfinder.VideoLinkBookingOptionsFinder
+import uk.gov.justice.digital.hmpps.whereabouts.services.vlboptionsfinder.VideoLinkBookingSearchSpecification
+import java.time.LocalDate
+import java.time.LocalDateTime
+import java.time.LocalTime
+
+/**
+ * This test class uses Mockk because I couldn't find a way to get mockito or mockitokotlin2 to compare Sequences.
+ * Not even with a custom Argument matcher.
+ * Show me how to do it and I'll revert to Mockito.
```
mikehalmamoj, line 24: I think that's a fantastic excuse to try out mockk.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/vlboptionsfinder/VideoLinkBookingOptionsFinder.kt
bell-pa: Hmm.  It's a good check to have, but strictly speaking its not needed because the Intervals come from the incoming spec and that has validation it that would rule this out.  You're right though, I could add an isValid method to Interval.  Its already there on the IntervalValidator in the same file as Validator.

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/CourtIntegrationTest.kt
bell-pa: OK. Will do.

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/services/locationfinder/AppointmentLocationServiceTest2.kt
```diff
@@ -0,0 +1,300 @@
...
+ */
+class AppointmentLocationServiceTest2 {
```
sp-watson, line 26: I can't help (sorry) but maybe make the class name more descriptive - I guess in this case the name would include "Sequences"?

src/main/resources/application.yml
bell-pa: yep.  Will do.

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/services/locationfinder/AppointmentLocationServiceTest2.kt
bell-pa: yeah.  Forgot about that.  I think that once the new algorithm is in use this class will become the only AppointmentLocationServicesTest class, but that's a feeble excuse.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/vlboptionsfinder/VideoLinkBookingOptionsFinder.kt
```diff
@@ -0,0 +1,160 @@
...
+           * that previousEventTime (not the time of the current event):
+           */
+          updateStateForPreviousEventTime()
+        }
+        // Now update state from the current event.
+        currentBookings += when (it) {
+          is StartEvent -> 1
+          is EndEvent -> -1
+        }
+        previousEventTime = it.time
```
andrewrlee, line 79: Might find it interesting, I remembered I had a go at [writing an implementation](https://gist.github.com/andrewrlee/681f62f9cb27261e899bc79649d842f9) of this about 6 months ago, when we first started looking at availability which used a similar approach.

I keyed the timeline based on number of minutes of the day each block started and then flattened the appointments 

The general bit for creating a time line and modifying it for each booking: 
```
fun createTimeLine(startOfDay: LocalTime, endOfDay: LocalTime, interval: Long): TimeLine {
    val indices = (startOfDay.minutesOfDay()..endOfDay.minutesOfDay() step interval)
    return indices.map { Pair(it, mutableListOf<Room>()) }.toMap(TreeMap())
}

fun TimeLine.add(booking: RoomBooking) = (booking.start until booking.end).forEach { slot ->
    this.computeIfPresent(slot) { _, list -> list.also { it.add(booking.room) } }
}
```

Then could make a request for a given time range and see the rooms that were booked for a given time:

```
fun TimeLine.getBookedRoomsBetween(timeRange: TimeRange) = this.subMap(timeRange.start, timeRange.end).flatMap { it.value }.toSortedSet()

fun TimeLine.isAvailable(booking: RoomBooking) = booking.room !in this.getBookedRoomsBetween(booking.timeRange)
```

bell-pa: Using the map to sort the appointments is interesting. Looks like you create a map entry for each possible start time.  Not too bad when they're at 15 minute intervals. 
I suppose I could just add all the events to a map by event time, then work through the map from earliest to latest to find out how the events at each significant time changed availability.  That would be better I think.  Might do something like that if we need another variation later on.

andrewrlee (approved):
LGTM

PR #327, opened 2021-06-22
DCS-1042 new endpoint for getting rooms in prison

GurnankCheema:
The rooms must have an eventType of 'APP' and a locationType of 'VIDE'


bell-pa (approved):
LGTM

PR #328, opened 2021-06-23
DCS-1041 integration tests added

GurnankCheema:
For the prisonController. Previous test also tidied up so that both tests use the same implementation
of prisonApiMockServer.stubGetAgencyLocationsForTypeUnrestricted()


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/PrisonController.kt
```diff
@@ -7,15 +7,13 @@ import org.springframework.http.HttpStatus
...
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
-import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.ResponseStatus
 import org.springframework.web.bind.annotation.RestController
 import uk.gov.justice.digital.hmpps.whereabouts.services.LocationService
 import uk.gov.justice.digital.hmpps.whereabouts.services.locationfinder.LocationIdAndDescription
 
 @Api(tags = ["prison"])
 @RestController
```
andrewrlee, line 16: Can we rename this from PrisonController to Videolink booking rooms - it's fine for it to be its own resource.

You'd expect a PrisonController to return Prisons

GurnankCheema: renamed as follows
PrisonController -> VideoLinkBookingRoomsController
PrisonControllerTest -> VideoLinkBookingRoomsControllerTest
PrisonIntegrationTest -> VideoLinkBookingRoomsIntegrationTest

src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/VideoLinkBookingRoomsIntegrationTest.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    prisonApiMockServer.stubGetAgencyLocationsForTypeUnrestricted("MDI", "APP", getVideLocations())
+
+    webTestClient.get()
+      .uri("/video-link-rooms/MDI")
+      .headers(setHeaders())
+      .exchange()
+      .expectStatus().isOk
+      .expectBody()
+      .jsonPath("[0].locationId").isEqualTo(1)
```
petergphillips, line 20: this seems a bit naff - would be best to just assert the json since you are stubbing the response

petergphillips (approved):
LGTM

PR #332, opened 2021-06-28
DCS-1056 /video-link-rooms/{agencyId} returns user-friendly description

GurnankCheema:
The endpoint will return a user-friendly description of the video link booking room, is the description is available. Otherwise it will return the alternative, hyphenated description.


andrewrlee (approved):
LGTM

PR #338, opened 2021-07-01
DCS-1060 extending VLB update to accept changing courtId

louisenorris:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/dto/VideoLinkBookingUpdateSpecification.kt
```diff
@@ -5,6 +5,14 @@ import javax.validation.Valid
 
 data class VideoLinkBookingUpdateSpecification(
 
+  @ApiModelProperty(
+    value = "The identifier of the court that requires the appointment.  This must be one of the court identifier values from the courts register service.",
+    example = "CMBGMC",
+    required = false,
+    notes = "One of court or courtId must contain a value.  courtId takes precedence over court."
```
andrewrlee, line 12: Is this right? Isn't it always going to be courtId?

louisenorris: It is but wasn't that the breaking change we were talking about which we could update once BVL has been released if we decide to?

louisenorris: Updated and notes removed.

andrewrlee (approved):
LGTM

PR #339, opened 2021-07-02
DCS-1057 db migration script to assign court_id to booking events

GurnankCheema:
Assigning court_id will ensure CSV for reports and data studio will be more accurate.
There will be some anomalies where 'null' records will be in the outputs but that is because some records may have been created and then deleted.


andrewrlee (approved):
LGTM

PR #345, opened 2021-07-12
DCS-976 Minor refinement to video link booking event reporting

bell-pa:
Add a sort to the VideoLinkBookingEvent query. Ensures that values always come back in the same order.


GurnankCheema (approved):
LGTM

PR #352, opened 2021-08-18
Update to Moorland Whereabouts config

Ed209uk:
Moorland changed the name of one of their units (what was called the CASU is now the Segregation Unit). This is to update the Whereabouts config so it works for them.


JamieDraperUK (approved):

andrewrlee (approved):
LGTM

PR #359, opened 2021-10-13
ESWE-164 Endpoint to get UNACAB count data for a given offender …

steverendell:
… over a given amount of time.

A new endpoint for education which sits with other attendance-related endpoints.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.kt
+++ b/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.kt
@@ -338,12 +338,7 @@ class AttendanceService(
   ): MutableMap<YearMonth, AttendanceSummary> {
 
     val yearMonth = YearMonth.of(offenderAttendance.eventDate.year, offenderAttendance.eventDate.month)
-    var slot = map.get(yearMonth)
-    if (slot == null) {
-      val new = AttendanceSummary(yearMonth)
-      map.put(yearMonth, new)
-      slot = new
-    }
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.kt
```diff
@@ -329,6 +332,46 @@ class AttendanceService(
...
+    offenderAttendance: OffenderAttendance
+  ): MutableMap<YearMonth, AttendanceSummary> {
+
+    val yearMonth = YearMonth.of(offenderAttendance.eventDate.year, offenderAttendance.eventDate.month)
+    var slot = map.get(yearMonth)
+    if (slot == null) {
+      val new = AttendanceSummary(yearMonth)
+      map.put(yearMonth, new)
+      slot = new
+    }
```
andrewrlee, line 346: think you might be able to do? 

`val slot = map.getOrPut(yearMonth) { AttendanceSummary(yearMonth) }`

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.kt
```diff
@@ -329,6 +332,46 @@ class AttendanceService(
       }.toSet()
   }
 
+  private fun addAttendance(
+    map: MutableMap<YearMonth, AttendanceSummary>,
+    offenderAttendance: OffenderAttendance
+  ): MutableMap<YearMonth, AttendanceSummary> {
```
JamieDraperUK, line 338: Just checking, does the return type need to be Mutable?

PR #364, opened 2021-11-01
Nn 3739 investigate why the whereabouts api showed as down

richardpopple:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/health/HealthCheck.java
```diff
@@ -29,10 +29,10 @@ public Health health() {
                             .block(timeout);
             return Health.up().withDetail("HttpStatus", responseEntity.getStatusCode()).build();
         } catch (final WebClientResponseException e) {
-            log.error("Dependant health check failed", e);
+            log.error("Dependant health check failed in class: " + this.getClass(), e);
```
sp-watson, line 32: Not sure why you need this - won't the class always be HealthCheck?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/health/HealthCheck.java
```diff
@@ -29,10 +29,10 @@ public Health health() {
...
             return Health.down(e).withDetail("body", e.getResponseBodyAsString()).build();
         } catch (final Exception e) {
-            log.error("Dependant health check failed", e);
+            log.error("Dependant health check failed in class: " + this.getClass(), e);
```
sp-watson, line 35: Also, don't you want to distinguish between WebClient and other exceptions... just a slightly different log string?

sp-watson (approved):

PR #367, opened 2021-11-10
ESWE-169-detail-attendance-data

monskapinga:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/AttendancesController.kt
```diff
@@ -119,6 +120,22 @@ class AttendancesController(private val attendanceService: AttendanceService) {
     )
   }
 
+  @GetMapping("/offender/{offenderNo}/attendances-in-whereabouts-or-prisonapi")
```
steverendell, line 123: I think a simpler url would be better (this is exposing the implementation and it wont always be going to prison-api in the future)

lbennett-moj (approved):

PR #390, opened 2022-02-15
SDI-54: 🔒️  Bump versions for security issues

petergphillips:


build.gradle.kts
```diff
@@ -29,7 +29,6 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-oauth2-client")
 
   implementation("org.springframework:spring-webflux")
-  implementation("org.springframework.boot:spring-boot-starter-reactor-netty")
```
petergphillips, line 32: not needed as webflux includes it

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/PingEndpoint.kt
```diff
@@ -1,14 +0,0 @@
...
-import org.springframework.boot.actuate.endpoint.web.annotation.WebEndpoint
-import org.springframework.context.annotation.Configuration
-import org.springframework.http.MediaType.TEXT_PLAIN_VALUE
-
-@Configuration
-@WebEndpoint(id = "ping")
-class PingEndpoint {
-
-  @ReadOperation(produces = [TEXT_PLAIN_VALUE])
-  fun ping(): String = "pong"
```
petergphillips, line 13: I checked app insights and endpoint not used anymore

andymarke (approved):
LGTM

PR #403, opened 2022-03-24
SDI-129: ✨ Add in new absent sub reasons

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/AbsentReasonsController.kt
```diff
@@ -16,8 +26,22 @@ class AbsentReasonsController {
...
     AbsentReasonsDto(
-      AbsentReason.paidReasons,
-      AbsentReason.unpaidReasons,
-      AbsentReason.iepTriggers
+      paidReasons,
+      unpaidReasons,
+      iepTriggers,
+    )
+
+  @GetMapping(value = ["/v2"])
```
petergphillips, line 34: for now is new endpoint - will get DPS to switch over to using the endpoint, then remove old and rename v2 -> v

andymarke (approved):
LGTM

PR #407, opened 2022-03-24
pgp SDI 129 absent sub reason 4

petergphillips:
- SDI-129: ✨ Write sub reason to comment field when writing back to nomis
- SDI-129: ✨ Write sub reason to case note field too when writing case notes


andymarke (approved):
LGTM

PR #412, opened 2022-03-30
SDI-53: 💄 Remove added as will have to be added by front end instead if required

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/model/AbsentReason.kt
```diff
@@ -12,7 +12,7 @@ enum class AbsentReason(val label: String) {
   RefusedIncentiveLevelWarning("Refused to attend");
 
   val labelWithWarning: String
-    get() = if (iepTriggers.contains(this)) "$label - incentive level warning added" else label
+    get() = if (iepTriggers.contains(this)) "$label - incentive level warning" else label
```
petergphillips, line 15: turns out front end sometimes needs `added` and other times does not, so better to take off and add if required

hyderkhanmoj (approved):
LGTM

PR #413, opened 2022-03-31
SDI-53: ✨ Rename unacceptable absence (don't merge)

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/model/AbsentReason.kt
```diff
@@ -1,15 +1,15 @@
...
-  NotRequired("Not required to attend"),
-  Refused("Refused to attend"),
-  RefusedIncentiveLevelWarning("Refused to attend");
+enum class AbsentReason(val label: String, val eventOutcome: String) {
+  ApprovedCourse("Approved course", "ACCAB"),
+  AcceptableAbsence("Acceptable absence", "ACCAB"),
+  SessionCancelled("Session cancelled", "CANC"),
+  RestInCellOrSick("Rest in cell or sick", "REST"),
+  RestDay("Rest day", "REST"),
+  UnacceptableAbsenceIncentiveLevelWarning("Unacceptable absence", "UNACAB"),
```
petergphillips, line 9: we're going to add a new `UnacceptableAbsence` that doesn't have a warning, so rename this one prior to adding the new one.

src/main/resources/db/migration/V40__migrate_UnacceptableAbsence.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE OFFENDER_ATTENDANCE
+    ALTER COLUMN ABSENT_REASON TYPE VARCHAR(40);
```
petergphillips, line 2: have to bump column length to cope with rename

src/main/resources/db/migration/V40__migrate_UnacceptableAbsence.sql
```diff
@@ -0,0 +1,6 @@
...
+
+UPDATE OFFENDER_ATTENDANCE
+SET ABSENT_REASON = 'UnacceptableAbsenceIncentiveLevelWarning'
```
petergphillips, line 5: we'll be adding a new `UnacceptableAbsence` reason shortly, so hence have to rename this one

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/IEPWarningService.kt
```diff
@@ -87,9 +85,6 @@ class IEPWarningService(
 
   private fun formatReasonAndComment(reason: AbsentReason, subReason: AbsentSubReason?, comment: String?): String {
     val caseNoteComment = subReason?.let { "${subReason.label}. $comment" } ?: comment
-    return when (reason) {
-      AbsentReason.RefusedIncentiveLevelWarning -> "Refused - Incentive Level warning - $caseNoteComment"
```
petergphillips, line 91: This didn't really make much sense previously since the unacceptable absence also generated an incentive level warning but the code only put the comment on the refused.  I guess this was because there were two refused scenarios.  Anyway since the reason now knows how to format itself with a warning then better to just use that.

andymarke (dismissed):
LGTM

hyderkhanmoj (approved):
LGTM

PR #419, opened 2022-04-12
SDI-53: ✨ Also rename attendance change to match

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/model/AttendanceChangeValues.kt
```diff
@@ -6,7 +6,7 @@ enum class AttendanceChangeValues {
   SessionCancelled,
   RestInCellOrSick,
   RestDay,
-  UnacceptableAbsence,
+  UnacceptableAbsenceIncentiveLevelWarning,
```
petergphillips, line 9: really don't understand why this is a separate enum, but have now added a test to ensure that they match

richardpopple (approved):
LGTM

PR #421, opened 2022-04-13
NN-4056: Use latest spring boot plugin to fix security issues

sp-watson:


build.gradle.kts
```diff
@@ -1,7 +1,7 @@
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "4.1.3-beta"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "4.1.4-beta"
   kotlin("plugin.spring") version "1.6.10"
```
petergphillips, line 5: `4.1.4` bumps kotlin too so this should really go to `1.6.20` as well

sp-watson: Done

timoj58 (approved):

PR #424, opened 2022-04-19
SDI-126: ✨ Move attended stat up a level

petergphillips:


src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceStatisticsTest.kt
```diff
@@ -130,6 +130,7 @@ class AttendanceStatisticsTest {
     val stats = service.getStats(prisonId, period, from, to)
 
     assertThat(stats).extracting("paidReasons").extracting("attended").isEqualTo(1)
```
petergphillips, line 132: leave old `attended` in place until front end starts using new one

steverendell (approved):

PR #441, opened 2022-05-23
SDI-212: ✨ New alternate endpoint to count activities

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceStatistics.kt
```diff
@@ -99,6 +99,41 @@ class AttendanceStatistics(
     )
   }
 
+  fun getStats2(prisonId: String, period: TimePeriod?, from: LocalDate, to: LocalDate): Stats {
+    val periods = period?.let { setOf(it) } ?: setOf(TimePeriod.PM, TimePeriod.AM)
+    val counts = prisonApiService.getScheduleActivityCount(prisonId, from, to, periods)
```
petergphillips, line 104: new endpoint that calls new Prison API endpoint to retrieve stats

steverendell (approved):
lgtm

PR #445, opened 2022-05-25
SDI-212: 🐛 Use new Prison API endpoint

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceStatistics.kt
```diff
@@ -101,20 +101,21 @@ class AttendanceStatistics(
...
+    val counts = prisonApiService.getScheduleActivityCounts(prisonId, from, to, periods, attendancesBookingIdsCount)
+
     val (paid, unpaid) = attendances.partition { AbsentReason.paidReasons.contains(it.absentReason) }
     val paidCounts = paid.groupBy { it.absentReason }.mapValues { it.value.count() }
     val unpaidCounts = unpaid.groupBy { it.absentReason }.mapValues { it.value.count() }
 
     return Stats(
       scheduleActivities = counts.total,
-      notRecorded = counts.total - attendances.size,
+      notRecorded = counts.notRecorded,
```
petergphillips, line 118: use counts value from Prison API as takes into account prisoners who have been released etc.

steverendell (approved):
lgtm

PR #447, opened 2022-05-31
SDI-213: ✨ New endpoint for prisoners unaccounted for

petergphillips:
At present DPS fires off two calls - one to Prison API to get the scheduled activities and another to Whereabouts to get the attendances.  Whereabouts then calls Prison API to get the scheduled activities in order to return the offender details for the attendances.  This means that the same Prison API endpoint is called by both.  DPS then takes the difference between both lists to find out the prisoners that are unaccounted for - those that haven't attended.

This change then adds a new endpoint to return the list of prisoners that are unaccounted for, by doing the difference in Whereabouts instead.  It calls the same endpoint in Prison API.

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/dto/PrisonerScheduleDto.kt
+++ b/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/dto/PrisonerScheduleDto.kt
@@ -1,0 +1,100 @@
+package uk.gov.justice.digital.hmpps.whereabouts.dto
+
+import com.fasterxml.jackson.annotation.JsonInclude
+import io.swagger.annotations.ApiModel
+import io.swagger.annotations.ApiModelProperty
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/dto/PrisonerScheduleDto.kt
```diff
@@ -0,0 +1,100 @@
...
+
+  @Schema(
+    required = true,
+    description = "The number which (uniquely) identifies the internal location associated with the Scheduled Event (Prisoner Schedule)"
+  )
+  val locationId: @NotNull Long?,
+
+  @Schema(required = true, description = "Offender first name")
+  val firstName: @NotBlank String?,
+
```
lbennett-moj, line 40: lots of nullables, but as its a copy.....

lbennett-moj (approved):
LGTM

PR #449, opened 2022-06-06
SDI-214: ✨ New endpoint for activities for absent reason

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/ControllerAdvice.kt
```diff
@@ -161,7 +161,7 @@ class ControllerAdvice {
 
   @ExceptionHandler(Exception::class)
   fun handleException(e: Exception): ResponseEntity<ErrorResponse> {
-    log.error("Unexpected exception {}", e.message)
+    log.error("Unexpected exception", e)
```
petergphillips, line 164: want the full stack trace if unexpected exception otherwise can't find the cause

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/ControllerAdvice.kt
```diff
@@ -175,7 +175,7 @@ class ControllerAdvice {
 
   @ExceptionHandler(ValidationException::class)
   fun handleValidationException(e: Exception): ResponseEntity<ErrorResponse> {
-    log.error("Validation exception {}", e)
+    log.error("Validation exception {}", e.message)
```
petergphillips, line 178: don't need it for validation exception though, this was a mistake highlighted by intellij as there is a `{}` param

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/PrisonApi.java
```diff
@@ -166,6 +166,18 @@ public List<OffenderDetails> getScheduleActivityOffenderData(final String prison
                 .bodyToMono(responseType)
                 .block();
     }
+    public List<OffenderDetails> getScheduleActivityOffenderData(final String prisonId,
+                                                                 final Set<Long> eventIds) {
+        final var responseType = new ParameterizedTypeReference<List<OffenderDetails>>() {
+        };
+
+        return webClient.post()
+                .uri("/schedules/{prisonId}/activities-by-event-ids", prisonId)
```
petergphillips, line 175: this is an existing endpoint in Prison API that goes to get the selected events

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.kt
```diff
@@ -306,6 +306,28 @@ class AttendanceService(
...
+  ): List<AbsenceDto> {
+
+    val periods = period?.let { setOf(it) } ?: setOf(TimePeriod.PM, TimePeriod.AM)
+    val endDate = toDate ?: fromDate
+
+    val attendances = attendanceRepository
+      .findByPrisonIdAndEventDateBetweenAndPeriodInAndAbsentReason(prisonId, fromDate, endDate, periods, absentReason)
+    val attendanceMap = attendances.associateBy({ it.eventId }, { it })
+
+    val offenderDetails = if (attendanceMap.isNotEmpty()) {
```
petergphillips, line 324: previous version (above) filtered by attendances anyway, so new version just gets the attendances that we know about rather than getting all attendances for all reasons and filtering

steverendell (approved):
lgtm

PR #450, opened 2022-06-06
SDI-214: 🐛 Fix whereabouts null pointer as time period not coming through from Prison API

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AttendanceService.kt
```diff
@@ -451,6 +451,26 @@ class AttendanceService(
...
 
+  private fun toAbsenceDto2(details: OffenderDetails, attendance: Attendance): AbsenceDto =
+    AbsenceDto(
+      attendanceId = attendance.id,
+      bookingId = attendance.bookingId,
+      offenderNo = details.offenderNo,
+      eventId = attendance.eventId,
+      eventLocationId = attendance.eventLocationId,
+      eventDate = attendance.eventDate,
+      period = attendance.period,
```
petergphillips, line 462: copied to pull the `period` through from the attendance instead.  Didn't want to alter the original as still in use, but will be removed shortly after testing new endpoint

mikehalmamoj (approved):

PR #456, opened 2022-06-10
SDI-212: ♻️  Switch old endpoint to use new code

petergphillips:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/AttendanceStatisticsController.kt
```diff
@@ -57,5 +57,5 @@ class AttendanceStatisticsController(private val attendanceStatistics: Attendanc
       description = "To date of event in format YYYY-MM-DD",
       required = true
     ) @RequestParam(name = "toDate") @DateTimeFormat(iso = DATE) toDate: LocalDate
-  ): Stats = attendanceStatistics.getStats2(prisonId, period, fromDate, toDate)
+  ): Stats = attendanceStatistics.getStats(prisonId, period, fromDate, toDate)
```
petergphillips, line 60: Change will be in two parts - this change makes `over-date-range` method now use the new calculations.  Second part will be getting DPS to call the `over-date-range` endpoint instead of `over-date-range-2`

PR #468, opened 2022-07-25
NN-4278 generic service upgrade

timoj58:


timoj58:
i think i may need to upgrade the README in helm as well?

helm_deploy/whereabouts-api/values.yaml
```diff
@@ -1,18 +1,54 @@
...
+  image:
+    repository: quay.io/hmpps/whereabouts-api
+    tag: latest
+    port: 8080
+
+  ingress:
+    enabled: true
+    host: app-hostname.local
+    tlsSecretName: whereabouts-api-cert
+    contextColour: green
```
mikehalmamoj, line 14: This isn't required anymore and doesn't do anything - it can be removed from those projects it exists in.

helm_deploy/whereabouts-api/values.yaml
```diff
@@ -1,18 +1,54 @@
 ---
 # Values here are the same across all environments
+generic-service:
```
mikehalmamoj, line 3: Hmmm lots of other projects seem to set `nameOverride` here - but it [defaults](https://github.com/ministryofjustice/hmpps-helm-charts/blob/654ca0e77a74eb847c0cd38dbaa6eb7492daa4e1/charts/generic-service/templates/_helpers.tpl#L6) to `.Chart.Name` so I think you're right to leave the default. We'll find out on deployment!

mikehalmamoj (approved):

mikehalmamoj:
> i think i may need to upgrade the README in helm as well?

Ah yes, tiller is long gone. If you need the latest helm commands for history and rollback they will be similar to some documentation I wrote [here](https://github.com/ministryofjustice/send-legal-mail-to-prisons-api#rolling-back-the-application).

mikehalmamoj (approved):

PR #472, opened 2022-07-27
NN-4278 generic service upgrade 

timoj58:
this is previous branch, now that ingress has been updated


sp-watson (approved):
Apparently same as previously approved

PR #476, opened 2022-07-27
NN-4278 add v1.2 ingress

timoj58:


helm_deploy/whereabouts-api/values.yaml
```diff
@@ -8,7 +8,7 @@ ingress:
   # live-1 is blue, live is green
   contextColour: green
   annotations:
-    external-dns.alpha.kubernetes.io/aws-weight: "100"
+    external-dns.alpha.kubernetes.io/aws-weight: "0"
```
sp-watson, line 11: Doesn't this need to be kept at 100?

helm_deploy/whereabouts-api/templates/ingress-v1-2.yaml
```diff
@@ -0,0 +1,36 @@
+{{- if .Values.ingress.enabled -}}
```
sp-watson: Not sure whether this will work - can't really review it

timoj58: its just coped from the other ingress.  its same value

helm_deploy/whereabouts-api/values.yaml
timoj58: the old one is set to 0 (which is a valid value) and the new to 100 as per the guide

sp-watson (approved):

PR #491, opened 2022-09-05
DCS-1833 Adding new courts

andrewrlee:
Also fixed the previous latest migration as the file name structure lacked double underscore meant that it was never applied.


mgreen-mojhub (approved):
LGTM

petergphillips (approved):
LGTM

PR #499, opened 2022-09-16
DCS-1847 tests refactoring

michalnawrockidj:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/controllers/AbsentReasonsController.kt
```diff
@@ -29,7 +29,7 @@ class AbsentReasonsController {
     iepTriggers.toList(),
     absentSubReasonTriggers,
     paidSubReasons.toSubReasonDto(),
-    unpaidSubReasons.toSubReasonDto(),
```
andrewrlee, line 32: think the trailing comma rule should be applied? 

Did you run ktlintFormat?

PR #508, opened 2022-10-19
DCS-1819 Fixing issue when deleting appointments

andrewrlee:
and no user token available

This matches the pattern used by the attendance service

### DIFF
--- a/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/config/VideoLinkBookingServiceForAppScope.kt
+++ b/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/config/VideoLinkBookingServiceForAppScope.kt
@@ -1,0 +1,38 @@
+package uk.gov.justice.digital.hmpps.whereabouts.config
+
+import org.springframework.beans.factory.annotation.Qualifier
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/config/VideoLinkBookingServiceForAppScope.kt
```diff
@@ -0,0 +1,38 @@
...
+import uk.gov.justice.digital.hmpps.whereabouts.services.court.VideoLinkBookingEventListener
+import uk.gov.justice.digital.hmpps.whereabouts.services.court.VideoLinkBookingService
+import java.time.Clock
+
+@Configuration
+class VideoLinkBookingServiceForAppScope {
+  @Bean(name = ["videoLinkBookingServiceAppScope"])
+  fun getPrisonServiceForAppScope(
+    courtService: CourtService,
+    @Qualifier("elite2WebClientAppScope") webClient: WebClient,
```
andrewrlee, line 21: needed to use a token with App Scope as the non app scope relies on a user token to be present

michalnawrockidj (approved):
LGTM

PR #509, opened 2022-10-20
DCS-1888 Prevent event generation when BVL makes requests

andrewrlee:
BVL consumes events regarding appointments but is also sometimes the source of the event.

When BVL is the source of an event, there is a race conditions, where the event listener processes the event before BVL has committed its transaction which produced the event.
 
This is a bit of hack maybe, that will prevent raising events for appointments that BVL is modifying.
It has the knock on effect that other systems will not be notified of these


louisenorris (approved):
LGTM

davidthomas22 (approved):
LGTM

PR #519, opened 2022-11-17
DCS-1913 removing unused endpoint

michalnawrockidj:


michalnawrockidj:
Please do not merge this to main branch until I give the green light

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/PrisonApi.java
```diff
@@ -358,20 +358,6 @@ public List<ScheduledAppointmentSearchDto> getScheduledAppointments(final String
                 .block();
     }
 
-
-    public List<ScheduledAppointmentDto> getScheduledAppointments(final String agencyId, final LocalDate date) {
```
louisenorris, line 362: Assume if this is being removed that nothing else is using it?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/VideoLinkBookingMigrationService.kt
```diff
@@ -57,11 +57,11 @@ class VideoLinkBookingMigrationService(
       videoLinkBooking.prisonId = nomisMainAppointment.agencyId
       videoLinkBooking.comment = nomisMainAppointment.comment
 
-      if (preAppointment != null && preAppointment?.let { updateAppointment(it) } == null) {
+      if (preAppointment != null && updateAppointment(preAppointment) == null) {
         videoLinkBooking.appointments.remove(HearingType.PRE)
         outcomes.add(PRE_DELETED)
       }
-      if (postAppointment != null && postAppointment?.let { updateAppointment(it) } == null) {
```
louisenorris, line 64: Do we def not need `preAppointment?.let` and `postAppointment?.let` anymore? Appriciate it looks like we check the same thing twice with not equal to null but maybe this is checking for another falsely value like undefined?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/PrisonApi.java
michalnawrockidj: Yes it is not used

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/VideoLinkBookingMigrationService.kt
michalnawrockidj: This is a small code improvement not related to actual code cleaning.   
postAppointment?.let { updateAppointment(it) } execute updateAppointment when postAppointment is not null however we added postAppointment != null what means in case postAppointment is null we never reach this point postAppointment?.let { updateAppointment(it) } it means we do not need to do this check again.

louisenorris (approved):
LGTM

connormaglynn (approved):
lgtm as well 😁 🚀

PR #520, opened 2022-11-23
DCS-1921 fixing PRE and POST appointment deletion

michalnawrockidj:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/model/VideoLinkBooking.kt
```diff
@@ -25,12 +25,12 @@ class VideoLinkBooking(
...
   val madeByTheCourt: Boolean? = true,
   var prisonId: String,
-  var comment: String? = null,
+  var comment: String? = null
 
 ) : BaseEntity(id) {
   @OneToMany(
     mappedBy = "videoLinkBooking",
-    fetch = FetchType.EAGER,
+    fetch = FetchType.LAZY,
```
connormaglynn, line 33: Changing this to lazy prevents us from pulling through the Bookings when searching for Appointments.

I believe the effect of this is that Bookings is not saved in the EntityManager state, therefor not allowing us to just delete Appointments without updating the state for Bookings.

Not entirely sure about the knock on effect of this, though.

PR #532, opened 2023-02-09
DCS-2027-bumping version

michalnawrockidj:


build.gradle.kts
```diff
@@ -1,7 +1,7 @@
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "4.8.3"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "5.0.1"
```
petergphillips, line 4: latest version is `5.1.0`, with a new `5.1.1-beta` out this morning

petergphillips:
There is a bug in the latest app insights which causes `InProc` requests to spam the logs - so you’ll need the sampling overrides for your two json configuration files like in https://github.com/ministryofjustice/prison-offender-events/blob/main/applicationinsights.json
Logback needs changes as the `ApplicationInsightsAppender` isn’t included anymore - you only need the `consoleAppender` as in https://github.com/ministryofjustice/hmpps-restricted-patients-api/blob/1baeb01f93[…]b16625f9b92579649ee3ca600/src/main/resources/logback-spring.xml
You seem to have deleted the `ClientTrackingConfigurationTest.kt` - would be better to copy from [https://github.com/ministryofjustice/hmpps-restricted-patients-api/blob/1baeb01f93[…]restrictedpatientsapi/config/ClientTrackingConfigurationTest.kt](https://github.com/ministryofjustice/hmpps-restricted-patients-api/blob/1baeb01f93d1455b16625f9b92579649ee3ca600/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsrestrictedpatientsapi/config/ClientTrackingConfigurationTest.kt) or similar

petergphillips (approved):
LGTM

PR #533, opened 2023-02-13
DCS-2013 add court type table and values for VLBs

louisenorris:


louisenorris:
Using a different approach

PR #540, opened 2023-02-15
DCS-2013 add court type for VLBs

louisenorris:


andrewrlee (approved):
LGTM

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/dto/CourtHearingTypeDto.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.whereabouts.dto
+
+import com.fasterxml.jackson.annotation.JsonInclude
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+data class CourtHearingTypeDto(
```
andrewrlee, line 6: you could probably remove the need for this class and the `map`

Jackson uses annotations to alter the format of the produced json.

You could use `@JsonFormat` to render enums as objects:
https://stackoverflow.com/a/69917399 

that should produce something like: 
`{ "name": "APPEAL", "description": "Appeal"}`
(Java enums have a name attribute that is the name of the enum instance itsself)

To prevent having the pass entire objects when setting the value, you should be able to use the JsonCreator annotation

louisenorris: Thanks, will address this in a follow up PR.

PR #569, opened 2023-06-12
HEAT-74: update to use new service account from latest irsa terraform module

neilmendum:
https://dsdmoj.atlassian.net/browse/HEAT-74

### DIFF
--- a/helm_deploy/whereabouts-api/values.yaml
+++ b/helm_deploy/whereabouts-api/values.yaml
@@ -4,6 +4,8 @@ generic-service:
 
   replicaCount: 4
 
+  serviceAccountName: whereabouts-api
+
   image:
     repository: quay.io/hmpps/whereabouts-api
     tag: app_version    # override at deployment time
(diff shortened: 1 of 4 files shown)


helm_deploy/whereabouts-api/values.yaml
```diff
@@ -4,6 +4,8 @@ generic-service:
 
   replicaCount: 4
 
+  serviceAccountName: whereabouts-api
```
neilmendum, line 7: Created with
- https://github.com/ministryofjustice/cloud-platform-environments/pull/13794/files#diff-ddd0f5c08baa00f066299c1e60f0720663cf02e8a43de742300f31985c8cf95dR17
- https://github.com/ministryofjustice/cloud-platform-environments/pull/13796/files#diff-3c4b2b0c21a617ecfe653c5e0e5db526e57e38fd4d7169332cc691a0c17ed9e4R17
- https://github.com/ministryofjustice/cloud-platform-environments/pull/13797/files#diff-8be730931e74fe680e079d6d9de0f9a3c3f2ee28690396b0671dfc34b7fa456aR17

As the IRSA module v2 service names only need to be unique within a namespace, rather than the cluster

PR #584, opened 2023-08-07
SAA-1060: Correction for Risley cell locations.

timharrison-moj:
Correction to show cell 20 on the A-wing All landings sub-location.


steverendell (approved):
lgtm

PR #599, opened 2023-09-28
Fixing logging when dev profile is switched on

andrewrlee:


src/main/resources/logback-spring.xml
```diff
@@ -6,22 +6,11 @@
 
     <springProperty scope="context" name="app" source="spring.application.name"/>
 
-    <springProfile name="dev || stdout">
```
andrewrlee, line 9: these tests were surely the wrong way round?

src/main/resources/logback-spring.xml
```diff
@@ -6,22 +6,11 @@
...
-            <layout class="ch.qos.logback.classic.PatternLayout">
-                <Pattern>${LOG_PATTERN}</Pattern>
-            </layout>
-        </appender>
-    </springProfile>
+    <appender name="consoleAppender" class="ch.qos.logback.core.ConsoleAppender">
+        <layout class="ch.qos.logback.classic.PatternLayout">
+            <Pattern>${LOG_PATTERN}</Pattern>
+        </layout>
+    </appender>
```
andrewrlee, line 13: just always use console STD out logging

jimbali (approved):

PR #603, opened 2023-10-03
MAP-253: Add eventId to BVL CSV

jimbali:
The eventId was not being copied over properly so it was missing from the CSV.


GurnankCheema (approved):
LGTM

PR #608, opened 2023-10-27
SAA-1343: Adding Ranby House block 8 as a new location for unlock

timharrison-moj:
Please see Eddie for reasons - new locations at Ranby.


petergphillips (approved):
LGTM

PR #618, opened 2023-12-07
Map 483 domain events sqs

michalnawrockidj:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/listeners/Message.kt
```diff
@@ -0,0 +1,55 @@
...
+  val reason: Reason,
+  val prisonId: String,
+)
+
+data class ReleasedOffenderEventMessage(
+  val occurredAt: String,
+  val additionalInformation: AdditionalInformation,
+)
+enum class ScheduleEventStatus {
+  CANC, COMP, EXP, SCH
```
petergphillips, line 45: not a great fan of having these two as enums.  I think it will mean that if there is a new event type then the code will fall over and message will stay on the DLQ.  If that's what you want then I guess fine

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/listeners/SqsDomainEventListener.kt
```diff
@@ -0,0 +1,35 @@
...
+  @Qualifier("videoLinkBookingServiceAppScope")
+  private val videoLinkBookingService: VideoLinkBookingService,
+  private val gson: Gson,
+) {
+  companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  @SqsListener("domainevent", factory = "hmppsQueueContainerFactoryProxy")
+  fun handleDomainEvents(requestJson: String?) {
```
petergphillips, line 22: we normally add a `@WithSpan` here so that we get an event in app insights - see https://github.com/ministryofjustice/hmpps-prisoner-search/blob/bde8f3c3d11237afecd167a90b199ca919ad3c00/hmpps-prisoner-search-indexer/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonersearch/indexer/listeners/DomainEventListener.kt#L36

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/court/VideoLinkBookingService.kt
```diff
@@ -339,4 +343,27 @@ class VideoLinkBookingService(
...
+      )
+      if (offenderBookings.isNotEmpty()) {
+        val prisonerAppointments =
+          videoLinkAppointmentRepository.findAllByHearingTypeIsAndStartDateTimeIsAfterAndVideoLinkBookingOffenderBookingIdIsAndVideoLinkBookingPrisonIdIs(
+            hearingType = MAIN,
+            startDateTime = LocalDateTime.parse(
+              releasedOffenderEventMessage.occurredAt,
+              DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss'Z'"),
+            ),
+            offenderBookingId = offenderBookings.first().bookingId,
```
petergphillips, line 362: is this right? How do you know it will be the first booking returned?  Looking at https://github.com/ministryofjustice/prison-api/blob/17d18b986f66d7881fffe17ea2999860cfdd2f7b/src/main/java/uk/gov/justice/hmpps/prison/repository/sql/InmateRepositorySql.kt#L91 there is no sort order on the query

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/court/VideoLinkBookingService.kt
```diff
@@ -339,4 +343,27 @@ class VideoLinkBookingService(
       videoLinkBookingEventListener.appointmentUpdatedInNomis(videoLinkAppointment, appointmentChangedEventMessage)
     }
   }
+
+  fun deleteAppointments(releasedOffenderEventMessage: ReleasedOffenderEventMessage) {
+    val reason = releasedOffenderEventMessage.additionalInformation.reason
+    if (reason == Reason.TRANSFERRED || reason == Reason.RELEASED) {
+      val offenderBookings = prisonApiService.getOffenderDetailsFromOffenderNos(
+        listOf(releasedOffenderEventMessage.additionalInformation.nomsNumber),
+        false,
```
petergphillips, line 352: kotlin style here - would be better to name the parameter as `false` doesn't mean anything to me
```suggestion
        iaActive = false,
```

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/court/VideoLinkBookingService.kt
```diff
@@ -339,4 +343,27 @@ class VideoLinkBookingService(
       videoLinkBookingEventListener.appointmentUpdatedInNomis(videoLinkAppointment, appointmentChangedEventMessage)
     }
   }
+
+  fun deleteAppointments(releasedOffenderEventMessage: ReleasedOffenderEventMessage) {
+    val reason = releasedOffenderEventMessage.additionalInformation.reason
+    if (reason == Reason.TRANSFERRED || reason == Reason.RELEASED) {
```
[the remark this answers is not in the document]
GurnankCheema: we are only targeting transfers and complete releases. I guess with hosp, they will return

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/court/VideoLinkBookingService.kt
```diff
@@ -339,4 +343,27 @@ class VideoLinkBookingService(
       videoLinkBookingEventListener.appointmentUpdatedInNomis(videoLinkAppointment, appointmentChangedEventMessage)
     }
   }
+
+  fun deleteAppointments(releasedOffenderEventMessage: ReleasedOffenderEventMessage) {
```
petergphillips, line 347: might be better to clarify what this method is doing i.e. renamed to `maybeDeleteAppointments` since it doesn't always delete the appointments, or `deleteAppointmentsIfNoLongerInPrison`

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/listeners/SqsDomainEventListener.kt
GurnankCheema: I will add that. I can assume the value can be any reasonable string like 
'map-devs-hmpps_whereabouts_api_domain_queue' ?

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/court/VideoLinkBookingService.kt
GurnankCheema: the response from prisonApi should be a List of 1 item as we have a query string of activeOnly which we have set to true. 
/bookings/offenders?activeOnly={isActive}
This is based on the (reasonable?) assumption that there is only 1 active booking.

GurnankCheema (approved):
Reviewed by Michal Nawrocki
LGTM

PR #638, opened 2024-01-25
Update config file for Whereabouts to add K-Wing for HMP Wealstun

Ed209uk:
Adding K-Wing for HMP Wealstun in the config file


timharrison-moj (approved):
LGTM

PR #657, opened 2024-03-22
SAA-1535: BVLS alternative time view to not show times before 9am and after 5pm

smcveigh941:
The code changed is only consumed by one service which is book-a-video-link. Checked using app insights


timharrison-moj (approved):
LGTM

PR #682, opened 2024-06-04
MAP-1170 change exception handler to actual exception thrown

GurnankCheema:
Exception handing was failing (as consequently preventing the sqs message from being fully actioned and this looping into DLQ) because the incorrect exception type was being tested for. 
Actual exception thrown was WebClientResponseException not EntityNotFoundException


michalnawrockidj (approved):
LGTM

PR #690, opened 2024-07-16
BAVL 251 adding feature switch to allow event subscription to be turned off for old BVLS when new BVLS service is available.

matt-crowson-moj:
This change introduces a feature switch to turn off the listening to BVLS related events when the new replatformed BLVS service is made available in production.

When the new service is available it will be solely responsible for the processing of these events.

Note: this change also includes some unrelated linting fixes that showed up when compiling existing code (it was annoying me!)

### DIFF
--- a/helm_deploy/whereabouts-api/values.yaml
+++ b/helm_deploy/whereabouts-api/values.yaml
@@ -36,6 +36,7 @@ generic-service:
     APPLICATIONINSIGHTS_CONNECTION_STRING: "InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY)"
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
     HMPPS_SQS_USE_WEB_TOKEN: "true"
+    FEATURE_BVLS_ENABLED: "true"
 
   namespace_secrets:
     whereabouts-api:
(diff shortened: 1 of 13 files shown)


helm_deploy/whereabouts-api/values.yaml
```diff
@@ -36,6 +36,7 @@ generic-service:
     APPLICATIONINSIGHTS_CONNECTION_STRING: "InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY)"
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
     HMPPS_SQS_USE_WEB_TOKEN: "true"
+    FEATURE_BVLS_ENABLED: "true"
```
matt-crowson-moj, line 39: When the time comes this will be set to 'false' and a deployment will be needed.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/listeners/SqsDomainEventListener.kt
```diff
@@ -15,29 +16,36 @@ class SqsDomainEventListener(
...
+        when (eventType) {
+          "prison-offender-events.prisoner.released" -> {
+            val transferEventMessage = gson.fromJson(message, ReleasedOffenderEventMessage::class.java)
+            videoLinkBookingService.deleteAppointmentWhenTransferredOrReleased(transferEventMessage)
+          }
         }
+      } catch (e: Exception) {
+        log.error("processDomainEvent() Unexpected error", e)
+        throw e
       }
```
timharrison-moj, line 48: Did we want to log that the event has been received and ignored, or just not mention it?

matt-crowson-moj: I did ponder this very thought and went with no.  Happy to log if you feel strongly about it.  I did add a log message to the init so we at least know it is enabled or not.

timharrison-moj (approved):
LGTM

PR #695, opened 2024-08-12
Map 1480 whereabouts api remove location prefix endpoint

GurnankCheema:
remove GET /locations/{agencyId}/{group}/location-prefix endpoint no longer being used


GurnankCheema:
created new PR

PR #697, opened 2024-08-12
MAP-1464 add logs to help diagnose email bug

GurnankCheema:
Emails not being sent to one particular court when BVL appointments are being cancelled in Nomis


michalnawrockidj (approved):
LGTM

PR #700, opened 2024-08-13
Map 1480 whereabouts api delete endpoint

GurnankCheema:
Delete delete /locations/groups/{agencyid}/ {name} endpoint as now this data is being provided by locations-inside-prison-api


GurnankCheema:
New PR created

PR #717, opened 2024-09-18
BAVL-42: Cater for cancelled bookings having their video_link_booking row not present

timharrison-moj:
When a video link booking is cancelled, the video link booking row is fully removed from the database.
This change will reconstruct the booking details from the events which remain.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/VideoLinkMigrateIntegrationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/VideoLinkMigrateIntegrationTest.kt
@@ -97,7 +97,7 @@ class VideoLinkMigrateIntegrationTest : IntegrationTest() {
         VideoLinkBookingEventType.CREATE,
         VideoLinkBookingEventType.DELETE,
       )
-      assertThat(courtCode).isEqualTo("CVNTCC")
+      assertThat(courtCode).isEqualTo("UNKNOWN")
       assertThat(madeByTheCourt).isTrue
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/whereabouts/integration/VideoLinkMigrateIntegrationTest.kt
```diff
@@ -97,7 +97,7 @@ class VideoLinkMigrateIntegrationTest : IntegrationTest() {
         VideoLinkBookingEventType.CREATE,
         VideoLinkBookingEventType.DELETE,
       )
-      assertThat(courtCode).isEqualTo("CVNTCC")
+      assertThat(courtCode).isEqualTo("UNKNOWN")
```
matt-crowson-moj, line 100: Not sure I understand why this is `UNKNOWN`.  We will will have this from the original create event won't we?

timharrison-moj: It's just the test data which sets up the event isn't putting in a court code - and when the event has nothing, it defaults it to UNKNOWN. I could put a court code here, if it makes you happy?

matt-crowson-moj: Just wanted to make sure I wasn't missing something is all. Not always easy to tell with git change history on PR's.

matt-crowson-moj (approved):
LGTM

PR #723, opened 2024-09-25
Revert "Revert "BAVL-41 missing domain events topic from config, needed so we can publish to it.""

matt-crowson-moj:
Reverts ministryofjustice/whereabouts-api#722


timharrison-moj (approved):
LGTM

PR #724, opened 2024-09-26
BAVL-41 move domain events topic subscription queue config to dev only, otherwise would impact release to higher environments.

matt-crowson-moj:
Having the SNS topic subscription in the overarching values.yaml file would potentially impact releasing to higher environments without the necessary cloud platform queue configuration in place.


timharrison-moj (approved):
LGTM - adding to preprod too?

matt-crowson-moj:
@timharrison-moj will add to preprod in a separate PR when the necessary cloud platform config has gone in.

PR #728, opened 2024-10-01
BAVL-43 enabled courts can also contain probation teams, we also need to check here to identify potential probation bookings when migrating.

matt-crowson-moj:
Checking the court name on the actual booking or event is not enough to identify potential probation bookings.

We also need to check the court id against those set up in the enabled courts if there is no court name on the booking itself.


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/court/VideoLinkMigrationService.kt
```diff
@@ -176,6 +182,10 @@ class VideoLinkMigrationService(
     )
   }
 
+  private fun VideoLinkBooking.isProbationBooking() =
+    courtName?.uppercase()?.contains("PROBATION") == true ||
+      courtService.getCourtNameForCourtId(courtId)?.uppercase()?.contains("PROBATION") == true
```
matt-crowson-moj, line 187: It might pay to have a little peak at this service method.

This also handles the situation where the `court id == 'other'` on bookings to migrate.  There is no court for `other`.

PR #730, opened 2024-10-04
SAA-2080: Filter attendance changes by prison id

munslowl:
UI is erroring (timing out) when fetching lots of changes.

Agreed that we add an index and filtering by agency id to API


dafhoustonmoj (approved):
LGTM

PR #761, opened 2024-12-16
BAVL-476 remove SNS subcription to domain events for BVLS code.

matt-crowson-moj:
This change removes domain event subscription for legacy BVLS code.

These domain events are now handled by the live re-platformed version of the service.

The cloud platform config will be removed after [here](https://github.com/ministryofjustice/cloud-platform-environments/pull/28431).  It is safer to remove the code first prior to cloud platform.


timharrison-moj (approved):
LGTM

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/listeners/SqsOffenderEventListener.kt
```diff
@@ -48,16 +39,6 @@ class SqsOffenderEventListener(
             bookingIds,
           )
         }
```
timharrison-moj, line 41: Did Jon B say to leave the prisoner deletion stuff for GDPR? I can't see it being used now, but not 100%

matt-crowson-moj: I haven't spoken to Jon.

timharrison-moj (approved):
LGTM. Just unsure about the GDPR prisoner delete event. I see you've left this in place, which is probably the safest thing.

src/main/java/uk/gov/justice/digital/hmpps/whereabouts/listeners/SqsOffenderEventListener.kt
matt-crowson-moj: Leaving this is as for now not really our responsibility.

PR #762, opened 2025-01-06
BAVL-478 remove legacy BVLS endpoints but leaving underlying service code in place to keeps PRs small in case anything needs reverting.

matt-crowson-moj:
Keeping the changes small for removal of legacy code in case anything is inadvertently broken.


timharrison-moj (approved):
LGTM

PR #766, opened 2025-01-14
BAVL-478 removing domain/entities for legacy BVLS code, no longer needed/used.

matt-crowson-moj:
The API is almost clear of all references to video link bookings.

There is a DTO which still uses a video link booking DTO, however this needs careful removal (if at all) as it is part of the existing public API.

The underlying database still stands as well.  Do we want this removing? @timharrison-moj


timharrison-moj (approved):
LGTM.
Sanity check - the locations stuff is not used in booking bulk appointments is it? This feature still exists for the "whereabouts" prisons not on A&A.

matt-crowson-moj:
> LGTM. Sanity check - the locations stuff is not used in booking bulk appointments is it? This feature still exists for the "whereabouts" prisons not on A&A.

@timharrison-moj I believe this to be fine.  The classes/files are not being used or referenced anywhere.  I do understand your concerns as a lot of code is being taken out.

PR #979, opened 2026-06-04
SAA-4338: Fix create appointment

munslowl:


src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/AppointmentService.kt
```diff
@@ -91,7 +90,7 @@ class AppointmentService(
   @Transactional
   fun createAppointment(createAppointmentSpecification: CreateAppointmentSpecification): List<CreatedAppointmentDetailsDto> {
     val appointmentCreated =
-      prisonApiServiceAuditable.createAppointments(makePrisonAppointment(createAppointmentSpecification))
+      prisonApiService.createAppointments(makePrisonAppointment(createAppointmentSpecification))
```
munslowl, line 93: Changing to `prisonApiService` uses `elite2WebClient` which is defined [here](https://github.com/ministryofjustice/whereabouts-api/blob/77b16017c7a9d6657a9abb906ac9a079a558958a/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/config/WebClientConfiguration.kt#L39) and forwards the jwt auth token, including the username, so the caseload can be determined by prison-api

mitchell-dsilva (approved):

